PR #2, opened 2016-03-01
Function excluding preprocessor directives

nkvaltine:
add preprocessor defines to exclude HPWHinit_file.  currently set to include full code

### DIFF
--- a/dev/HPWH.cc
+++ b/dev/HPWH.cc
@@ -1685,7 +1685,7 @@ int HPWH::checkInputs(){
   return returnVal;
 }
 
-
+#ifndef HPWH_ABRIDGED
 int HPWH::HPWHinit_file(std::string configFile){
   simHasFailed = true;  //this gets cleared on successful completion of init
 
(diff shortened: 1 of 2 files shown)


chipbarnaby:
Looks fine.

One C++ comment ... you can use

const int CONDENSITY_SIZE = 12;   // must be 12

instead of

-#define CONDENSITY_SIZE 12  /*\* this must be an integer, and only the value 12 

That enforces the type being integer.  Similarly const float etc. etc. for other constants.  Another advantage is then the constant value is visible in the debugger (at least on Visual Studio).

ghost:
These messages are still going into the info@ecotope.commailto:info@ecotope.com folder.  Can we change the settings?

Charlotte Trelease  | Office Manager
Ecotope, Inc. | 4056 9th Ave NE | Seattle, WA 98105
206.322.3753| Direct 206.596. 4718
www.ecotope.comhttp://www.ecotope.com/
[EcotopeLogoColorWebsite]
Monday 8-6
Tuesday 8-3
Wednesday 8-3
Friday 8-5
Out of the Office Thursday

PR #23, opened 2016-03-24
lockout on changing sanden setpoint

nkvaltine:


nkvaltine:
oops, forgot to increment version, let me do that before merging

PR #46, opened 2019-06-20
Update HPWH.cc for UA losses by node

pmskintner:
Changed the UA losses so it now uses the node temperatures instead of the average tank temperatures.

Attached is a couple example files with the reference files - new output files plotted for the temperature and it's first derivative. The top plot has the reference temperatures, and a scaled draws (black) and total input power (red). A back of the envelope calculation tells us that a 30 oC difference from the average tank temperature will result in a difference in the heat loss of 0.005 oC/min which is what we see in the first ~60 min of the test 30 file.

[ test30_File_Stiebel220e.csv .pdf](https://github.com/EcotopeResearch/HPWHsim/files/3311634/test30_File_Stiebel220e.csv.pdf)

[ test30_Preset_AOSmithHPTU80.csv .pdf](https://github.com/EcotopeResearch/HPWHsim/files/3311640/test30_Preset_AOSmithHPTU80.csv.pdf)


chipbarnaby (approved):
Looks fine as a first step.  Need to see current code.

nealkruis:
@chipbarnaby it's probably best to wait to approve the review until @pmskintner has finished his changes to the branch. @pmskintner just let us know when you're ready for us to review. We'll also see changes as they are made, and we can comment along the way, but we will wait until you are done before the official approval.

nealkruis:
Also, @pmskintner, you can start with a [draft pull request](https://github.blog/2019-02-14-introducing-draft-pull-requests/) that you mark later for "ready for review".

nealkruis:
@pmskintner also, please consider using more informative commit messages. Some people go way overboard with [best practices](https://github.com/erlang/otp/wiki/writing-good-commit-messages), but my general rule is:
- Make it a complete sentence
- Start it with a verb

For example, "Add node-by-node UA calculations." The default of "Update <filename>" doesn't give much insight into what's changing for when you need to reference back to a specific commit.

Hope that helps!

PR #47, opened 2019-07-11
 Add inversion mixing and internal conduction 

pmskintner:
To introduce more realism into HPWHsim mixing of temperature inversions and internal conduction between nodes is added. For details see the commit. 

Graphs below come are from the DHW_inv test in cse, showing the 6 thermocouples before and after the changes were made.

image

image


nealkruis:
@pmskintner looking at the charts, it looks like the inlet water in one case is 32 C, and the other is 25 C. Am I reading this correctly? Can you attach the CSE test file so I can see how things are changing?

pmskintner:
I didn't change anything in the tests, so the input should be 45F in both cases. Attached is a zip with the .cse test file and the output in .csv. Also below there's a plot where I added the inlet water temperature. The flat section around minute 800 is after the hot water is injected on bottom and the water heater is running the top resistance element. 

image


[DHW_INV.zip](https://github.com/EcotopeResearch/HPWHsim/files/3507120/DHW_INV.zip)

src/HPWH.in.hh
```diff
@@ -449,6 +458,8 @@ class HPWH {
 	/**< the setpoint of the tank  */
 	double *tankTemps_C;
 	/**< an array holding the temperature of each node - 0 is the bottom node, numNodes is the top  */
+	double *tempTemps;
```
nealkruis: Maybe come up with a better name? it is still in Celsius, so maybe "nextTankTemps_C"?

test/main.cc
```diff
@@ -263,12 +274,26 @@ int main(int argc, char *argv[])
       drStatus = HPWH::DR_ENGAGE;
     }
 
+	//// Let's create a feedback so the inlet temperature is the outlet temperature - 5 minutes
```
nealkruis: Remove these lines?

test/main.cc
```diff
@@ -263,12 +274,26 @@ int main(int argc, char *argv[])
...
+	//}
+
     // Run the step
     hpwh.runOneStep(allSchedules[0][i],    // Inlet water temperature (C)
                       GAL_TO_L(allSchedules[1][i]),          // Flow in gallons
                       airTemp2,  // Ambient Temp (C)
                       allSchedules[3][i],  // External Temp (C)
-                      drStatus, 1.0);    // DDR Status (now an enum. Fixed for now as allow)
+                      drStatus, // DDR Status (now an enum. Fixed for now as allow)
+					  1.0);    // Minutes per step
```
nealkruis: What's going on with the whitespace here?

src/HPWH.cc
```diff
@@ -391,13 +403,14 @@ int HPWH::runOneStep(double inletT_C, double drawVolume_L,
 		energyRemovedFromEnvironment_kWh += (setOfSources[i].energyOutput_kWh - setOfSources[i].energyInput_kWh);
 	}
 
+	// check for inverted temperature profile
+	if (doInversionMixing) 	mixTankInversions();
```
nealkruis: put on new line with scoping brackets.

src/HPWH.cc
```diff
@@ -1207,22 +1227,108 @@ void HPWH::updateTankTemps(double drawVolume_L, double inletT_C, double tankAmbi
...
-	//get average tank temperature
-	double avgTemp = 0;
-	for (int i = 0; i < numNodes; i++) avgTemp += tankTemps_C[i];
-	avgTemp /= numNodes;
+	// calculate conduction between the nodes AND heat loss by node with top and bottom having greater surface area.
+	// model uses explicit finite difference to find conductive heat exchange between the tank nodes with the boundary conditions
+	// on the top and bottom node being the fraction of UA that corresponds to the top and bottom of the tank.  
+	if (doConduction) {
+		// height estimate from Rheem along with the volume is used to get the radius and node_height
+		double height = 1.2; //meters
```
nealkruis: `static const double height = 1.2; // meters`

src/HPWH.cc
```diff
@@ -1207,22 +1227,108 @@ void HPWH::updateTankTemps(double drawVolume_L, double inletT_C, double tankAmbi
...
+		double rad = sqrt(tankVolume_L / (1000 * 3.14159 * height));
```
nealkruis: `const double rad...`. There are a lot of things here you can probably make constants (i.e., they don't change within the scope).

src/HPWH.cc
```diff
@@ -1207,22 +1227,108 @@ void HPWH::updateTankTemps(double drawVolume_L, double inletT_C, double tankAmbi
...
+		double node_height = height/numNodes;
+
+		// Get the "constant" tau for the stability condition and the conduction calculation
+		double tau = KWATER_WpermC / ( CPWATER_kJperkgC * 1000 * DENSITYWATER_kgperL * 1000 * (node_height * node_height) ) * minutesPerStep * 60.0;
+		if (tau > 0.5) msg("The stability condition for conduction has failed, these results are going to be interesting!\n");
```
nealkruis: new line + bracket for "if" statement

src/HPWH.cc
```diff
@@ -1207,22 +1227,108 @@ void HPWH::updateTankTemps(double drawVolume_L, double inletT_C, double tankAmbi
...
+			double standbyLosses_kJ = (tankUA_kJperHrC * Ar / numNodes * (tankTemps_C[i] - tankAmbientT_C) * (minutesPerStep / 60.0));
+			standbyLosses_kWh += KJ_TO_KWH(standbyLosses_kJ);
+
+			//The effect of standby loss on temperature in each node
+			tempTemps[i] -= standbyLosses_kJ / ((volPerNode_LperNode * DENSITYWATER_kgperL) * CPWATER_kJperkgC);
+		}
+	}
+	else { // Ignore tank conduction and calculate UA losses in the old way.
+		
+		for (int i = 0; i < numNodes; i++) 	tempTemps[i] = tankTemps_C[i];
```
nealkruis: new line + bracket for "if" statement (check throughout).

src/HPWH.cc
```diff
@@ -1207,22 +1227,108 @@ void HPWH::updateTankTemps(double drawVolume_L, double inletT_C, double tankAmbi
...
+			//faction of tank area on the sides
+			//kJ's lost as standby in the current time step for each node.
+			double standbyLosses_kJ = (tankUA_kJperHrC * Ar / numNodes * (tankTemps_C[i] - tankAmbientT_C) * (minutesPerStep / 60.0));
+			standbyLosses_kWh += KJ_TO_KWH(standbyLosses_kJ);
+
+			//The effect of standby loss on temperature in each node
+			tempTemps[i] -= standbyLosses_kJ / ((volPerNode_LperNode * DENSITYWATER_kgperL) * CPWATER_kJperkgC);
+		}
+	}
+	else { // Ignore tank conduction and calculate UA losses in the old way.
```
nealkruis: It's not clear we would want to ever use the old UA loss method. If we did, it probably shouldn't be compounded with how/if we are calculating conduction.

src/HPWH.cc
```diff
@@ -1207,22 +1227,108 @@ void HPWH::updateTankTemps(double drawVolume_L, double inletT_C, double tankAmbi
...
-	//The effect of standby loss on temperature in each segment
-	double lossPerNode_C = (standbyLosses_kJ / numNodes) / ((volPerNode_LperNode * DENSITYWATER_kgperL) * CPWATER_kJperkgC);
-	for (int i = 0; i < numNodes; i++) tankTemps_C[i] -= lossPerNode_C;
 }  //end updateTankTemps
 
 
+// Inversion mixing modeled after bigladder EnergyPlus code PK
+void HPWH::mixTankInversions() {
+	bool hasInversion;
+	double volumePerNode_L = tankVolume_L / numNodes;
```
nealkruis: `const`

src/HPWH.cc
```diff
@@ -2682,9 +2805,14 @@ int HPWH::HPWHinit_resTank(double tankVol_L, double energyFactor, double upperPo
...
 	double denominator = 67.5 * ((24.0 / 41094.0) - temp);
-	double UA_btuPerHrF = numerator / denominator;
-	tankUA_kJperHrC = UA_btuPerHrF * (1.8 / 0.9478);
+	tankUA_kJperHrC = UAf_TO_UAc(numerator / denominator); 
 
+	if (tankUA_kJperHrC < 0.) {
+		if (hpwhVerbosity >= VRB_reluctant && tankUA_kJperHrC < -0.1) {
+			msg("Computed tankUA_kJperHrC is less than 0, and is reset to 0.");
+		}
+		tankUA_kJperHrC = 0;
```
nealkruis: Generally, it's good practice to always add a decimal point to floats and doubles so you don't accidentally do something where the compiler mistakes it for an integer. E.g., `float x = 2/3` makes x equal 0.0 (should be `float x = 2./3.` instead).

src/HPWH.cc
```diff
@@ -2729,6 +2857,8 @@ int HPWH::HPWHinit_genericHPWH(double tankVol_L, double energyFactor, double res
 	tankTemps_C = new double[numNodes];
 	setpoint_C = F_TO_C(127.0);
 
+	tempTemps = new double[numNodes];
```
nealkruis: We should look into converting all the double[] arrays to std::vector to avoid potential memory problems.

src/HPWH.cc
```diff
@@ -2139,6 +2246,12 @@ int HPWH::checkInputs(){
 
 	}
 
+	//Check if the UA is out of bounds
+	if (tankUA_kJperHrC < 0.0) {
+		msg("The tankUA_kJperHrC must be above 0 for a HPWH, tankUA_kJperHrC is: %f  \n", tankUA_kJperHrC);
+		returnVal = HPWH_ABORT;
```
chipbarnaby, line 2395: Imprecise message -- test is < 0, but message says "must be above 0".

src/HPWH.cc
```diff
@@ -2151,6 +2264,7 @@ int HPWH::HPWHinit_file(string configFile){
 
 	//clear out old stuff if you're re-initializing
 	if (tankTemps_C != NULL) delete[] tankTemps_C;
+	if (tempTemps != NULL) delete[] tempTemps;
 	if (setOfSources != NULL) delete[] setOfSources;
```
chipbarnaby, line 2412: if (p != NULL) delete[] p -- the NULL test is not needed, delete[] does nothing when p == NULL.  However, depending on the situation, it is often desirable to add p = NULL after the delete[] to be absolutely sure p will not be delete[]ed again.  Or use stdlib smart pointers.

Also, there seem to be several places where these arrays are allocated or deleted.  Should there be functions such as AllocArrays() and DeleteArrays() that are called from the several places?  That approach makes changes easier/safer -- no need to review the entire file, just change the functions.

src/HPWH.in.hh
```diff
@@ -475,6 +486,12 @@ class HPWH {
   double member_inletT_C;
   double member_minutesPerStep;
 
+  bool doInversionMixing;
+  /**<  whether or not to model will mix temperature inversions in the tank  */
+
```
chipbarnaby, line 491: Comments of this sort are ambiguous.  Something like the following is better.
// iff true, model inversion mixing <whatever>

where iff = if and only if

chipbarnaby (approved):
Generally looks good.  Several picky comments interleaved below.

I realize the code was inherited from others, warts and all.  We should all keep discussing preferred style and try to leave the code in better shape each time it gets worked on.


General style -- put if stuff on separate line
  if (xxx)
      doWhatever()

nealkruis (changes requested):
Looking pretty good. A little clean up here and there.

src/HPWH.cc
pmskintner: There's many places these arrays are allocated and deleted, and I'm not exactly sure why for each case. I'll remove NULL tests, but going through the entire file can probably wait.

test/main.cc
pmskintner: Not sure exactly what you mean, I tried respacing it out if it was showing up poorly formatted. But if you hate seeing white I can remove it, I think it adds to readability.

nealkruis: I was just talking about how the last line is indented so much further. Sometimes things show up this way in GitHub if there isn't consistent use of tabs vs. spaces.

src/HPWH.cc
pmskintner: So I changed it, but it's pretty screwy. Because the top node loses heat faster than the other nodes it constantly creates inversions. I made two plots below, the first is temperature, note the green line is the top node, and the second is the output power and the hot water draws. This also seems to result in weird behavior for the compressor.

I'm fine leaving this in as a setting for the research build but just thought you should be warned.  

image

image

nealkruis: It seems like we should always do inversion mixing, too.

nealkruis (approved):

PR #48, opened 2019-08-15
Limit UA

pmskintner:
Add checks on the UA (0 - 500) and the EF (0 - .98) to abort HPWHsim if they're out of bounds.


pmskintner:
No max for UA now

nealkruis (approved):
@pmskintner Looks good. Be careful to provide more informative commit messages. "Update HPWH.cc" doesn't tell me much about what changes I'm looking at.

PR #55, opened 2019-12-05
setTankSize_adjustUA

pmskintner:
Wrote setTankSize_adjustUA() to change the UA of a tank if change the tank size, by assuming the U is unchanged and using the initialized volume the new volume based on the function getTankSurfaceArea(). 

getTankSurfaceArea() finds the SA given a volume based on data from 88 insulated storage tanks from a variety of manufacturers.

### DIFF
--- a/src/HPWH.cc
+++ b/src/HPWH.cc
@@ -737,6 +737,35 @@ int HPWH::setDoTempDepression(bool doTempDepress) {
 	return 0;
 }
 
+int HPWH::setTankSize_adjustUA(double HPWH_size_L){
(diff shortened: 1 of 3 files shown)


chipbarnaby (approved):
A general "DRY" comment about all the HPWH::setWhatever functions --

It would be shorter / clearer to use a default argument for units and eliminate the no-units variant --

int HPWH::setTankSize_adjustUA( double HPWH_size, UNITS units=UNITS_L)

pmskintner:
I've been trying to just keep the same format. I can go through and "DRY" up the new setter functions.

PR #56, opened 2019-12-11
Radius Calc and Units Fix

pmskintner:
- Fixed L_TO_GAL in getTankVolume()
- Fixed Issue where heat lost from top and bottom wasn't added to standby
- DRYed newer getters and setters so defualt args take units.
- Added SA and length UNITS
- Add calculation for tank radius in for use in UA losses
- Updated ref tests
- getTankSurfaceArea takes UNITS_FT2 or UNITS_M2


chipbarnaby (approved):
The DRYing could be much simpler with default arguments ...

in HPWH.hh --
getXXX( UNITS units=UNITS_L)  or whatever

Then no need to change existing no-arg calls.

In other words, get default and non-default behavior out of one variant of the function.

chipbarnaby:
I approved but note my comment.  Did not merge.

pmskintner:
That makes perfect sense Chip. I'll change them up right now.

pmskintner:
Hey Chip, do you want to review changes again?

chipbarnaby (approved):
Looks good.
I often show default arg value in a comment in the implementation file --

float getXXX( UNITS units /*=UNITS_FURLONGS*/)
{    return blah blah
}

pmskintner:
Slowly but surely I'll perfect HPWHsim. I'll add this in so it's clear and merge this.

PR #67, opened 2020-04-23
New Aspect Ratio Based Height and Radius Calculation

pmskintner:
New calculation of tank radius by assuming a constant aspect ratio between the entire unit including the jacket and the inner hot water tank. Aspect ratio was derived from 88 tanks on the market.
-Tank radius is found from tank volume and aspect ratio
-Height is found from aspect ratio and radius
-Surface area is found from aspect ratio and radius

Results in changes over 1 degF in the tests, due to an increase in the rate of conduction because of dramatic decrease in the calculated tank height. New height matches the surveyed tanks much better than the old method.


chipbarnaby (approved):
What is the range of tank sizes used to derive the aspect ratio?

In CBECC-Res, the HPWH defaults will be used for tanks of any size, including very large ones (hundreds of gallons).  Any issues?

In addition, should be be considering insulation thickness?  The surface area being calculated is for the tank only (actually for the water only).  The external heat transfer surface area is larger .

pmskintner:
The range of tank sizes is from  40 to 1530 gallons, so no issues. Although it does look like there's slight downward trend to the aspect ratio with volume such that we'd consistently be over predicting the aspect ratio at tank sizes > 1000 gallons. 

At some point we might want to consider insulation thickness, but this might be too much work to be worth it. It'd be reasonable to adjust the UA by the insulation thickness. But I don't think we want to model heat transfer through the insulation on the same minute by minute scale.

The HPWH tanks have been assuming that the heat transfer out of the water into the jacket and out of the jacket to the air are equal at every timestep, or that the jacket is infinitesimally thin. The UA then comes from lab measurements for each preset. Under the thin jacket assumption this UA would be applied to the volume of water and we'd scale it to any other volume of water.

PR #103, opened 2021-03-01
Set Defaults Function in Constructor

pmskintner:
Defaults moved into function called in constructor and out of initialization list. This is recalled during preset or file model initialization to avoid carrying over non default values for variables in the case of re-initialization of a preset on the same object.

### DIFF
--- a/src/HPWH.cc
+++ b/src/HPWH.cc
@@ -67,15 +67,17 @@ const std::string HPWH::version_maint = HPWHVRSN_META;
 
 //the HPWH functions
 //the publics
-HPWH::HPWH() :
-	simHasFailed(true), isHeating(false), setpointFixed(false), tankSizeFixed(true), canScale(false), hpwhVerbosity(VRB_silent),
-	messageCallback(NULL), messageCallbackContextPtr(NULL), numHeatSources(0),
-	setOfSources(NULL), tankTemps_C(NULL), nextTankTemps_C(NULL), doTempDepression(false), 
-	locationTemperature_C(UNINITIALIZED_LOCATIONTEMP),
-	doInversionMixing(true), doConduction(true), 
-	inletHeight(0), inlet2Height(0), fittingsUA_kJperHrC(0.),
-	prevDRstatus(DR_ALLOW), timerLimitTOT(60.), timerTOT(0.)
-{  }
+HPWH::HPWH() { setAllDefaults(); };
+
+void HPWH::setAllDefaults() {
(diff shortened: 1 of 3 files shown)


chipbarnaby (changes requested):
Now I think there are issues with the heap pointers (tankTemps_C etc.).  Unilaterally setting them to NULL in setAllDefaults is a memory leak on 2nd and subsequent calls.

The constructor should set them NULL (or nullptr) and setAllDefaults should delete[] them, then set them to nullptr.

The destructor should delete[] them. 

Note delete[] is a NOP if the ptr is null.

pmskintner:
> Now I think there are issues with the heap pointers (tankTemps_C etc.). Unilaterally setting them to NULL in setAllDefaults is a memory leak on 2nd and subsequent calls.
> 
> The constructor should set them NULL (or nullptr) and setAllDefaults should delete[] them, then set them to nullptr.
> 
> The destructor should delete[] them.
> 
> Note delete[] is a NOP if the ptr is null.

Just to make sure I understand. In the first call to a preset the heap pointers are set and point to a memory block. But as it stands when there's a second call if I just set the pointers to NULL then the memory blocks still exist, I've just set the pointer to null (so they're lost forever). And when I reassign tankTemps_C (etc.) I'm calling them with new so I've just created a whole new memory block.  

delete[] will obviously remove the memory block in the first place.

chipbarnaby:
Right.  "lost forever" = memory leak.  The memory is cleaned up when the
program closes, but it is lost during execution.  In egregious cases,
programs can crash cuz they run out of memory.

The memory blocks can be different sizes depending on the preset (right?).
The blocks cannot be resized after allocation, so the best simplest
strategy is to always delete[] them before presetting.

Or to be more modern, could convert them to std::vector.  Then they would
exist from constructor through destructor.  preset should probably reserve
sufficient space in each.  I think if this is done right, there would be no
add'l overhead.  Probably would involve a bunch of changes, but maybe not
since std::vector has operator[] that does the natural thing.

Chip

pmskintner:
> The memory blocks can be different sizes depending on the preset (right?).

Correct they're ranging from 12 - 96. 

I know the convert to vector is on the issue list, I was planning on getting to it when I have some free time... 

Did you see my [second commit](https://github.com/EcotopeResearch/HPWHsim/pull/103/commits/58ae6b2869cd237d029eb31bfbad3518e3a0b5e7)?

chipbarnaby (approved):
Looks good!

chipbarnaby:
Approved, did not merge.

PR #110, opened 2021-03-09
Cap Compressor Outlet Temperature

pmskintner:
- isNewSetpointPossible() returns if new setpoint is possible and the max allowed setpoint
  - True if has a compressor and the new setpoint is below the compressor max output.
  - True if there is a resistance element in the tank and the new setpoint is below 100C
  - True if it is a storage tank. These don't really operate on the concept of a setpoint anyways.

- Can set setpoint higher than the compressor's max iff the hpwh has resistance.
  - Idea is to heat up with compressor to the setpoint then switch to resistance. Sometimes preset logic overrides this if an element has relative turn on temperature, the new setpoint can be high enough from the old that the top element turns on.
  - If heat source ends early checks if backup needs to run iff the heat source is "maxed out". A compressor needs to run backup heat source after reaching max setpoint or it will spend the next minute getting backup to setpoint.
  
- Changes to reference tests from bug fixes:
  - Bug fix where inversions were not mixed in SP HPWH's when setpoint temperature was decreased. In case of cold weather packages the setpoint would be decreased when air temperature drops too low (i.e. < 40F)
  - Change reference test for CxV_5 with elevated temperature since it can't make that temperature! Leads to drastic change.

- Expand testing for max setpoints. 
- Minor simplification of presets.


chipbarnaby (approved):
Looks good, will modify setpoint code in CSE.

PR #111, opened 2021-03-09
check setpoint fixed in setpoint possible?

pmskintner:
isNewSetpointPossible() checks if setpoint is fixed.
 - fixes unit bug.


chipbarnaby (approved):
Looks OK, will test in CSE when merged.

PR #117, opened 2021-04-21
 Add tolerance to sizing fraction check

pmskintner:


chipbarnaby (approved):
I worry that although this change will prevent false messages, it leaves the door open for other problems when the values are in the wrong order, even by small amounts.  In other words, the fix hides the problem in this case, but the values remain wrong.

I would think it would be possible to force the values to be consistent when they are derived in getSizingFractions().

PR #131, opened 2021-09-09
Res Tank Generic with R-value

pmskintner:
- Change HPWHinit_commercialResTank() to HPWHinit_resTankGeneric().
  - Input for R value in IP units.
  - Replace commercial enums with MODELS_CustomResTankGeneric.


chipbarnaby (commented):
I don't understand the motivation for these changes.
1.  Is there still a way to create a EF (or UEF) based tank (e.g. commercially available < 120 gal).
2. Have the "special" swing tank controls been eliminated?
Will communicate by email.

PR #137, opened 2022-01-25
MP Generic

pmskintner:
- Add scalable generic MP heater: MODELS_Scalable_MP.
  - Scales with same functions for the scalable SP model.


chipbarnaby (approved):
I cannot review / comment on files that support testing etc.

PR #142, opened 2022-03-23
Add btwxt for Mitsu QAHV

pmskintner:
- Adds btwxt grid interpolation package for calculating performance in single pass and multipass HPWHs
- Adds new single pass HPWH model: MODELS_MITSUBISHI_QAHV_N136TAU_HPB_SP
- Converts Nyle MP models from exponential regression to btwxt grid interpolation.
  - Some changes in Nyle performance up to 5%
- Update regression tests for Nyle MP and new regression tests for QAHV.


chipbarnaby (changes requested):
Do I understand that in CSE there will be folders vendor\btwxt and vendor\hpwhsim\vendor\btwxt?

Is there logic to cause hpwhsim to use the already-present btwxt in the CSE context?

Semi-related, I would think it should be possible to hide most of btwxt within the source file so hpwh.h does not have to #include btwxt. 

For example,
You can probably use
class Btwxt::RegularGridInterpolator *whatever
rather than
Btwxt::RegularGridInterpolator *whatever

to avoid the need for the whole enchilada.

The idea is to make the publicly visible stuff as lean as possible.

vendor/CMakeLists.txt
```diff
@@ -0,0 +1,17 @@
...
+    if(GIT_SUBMODULE)
+        message(STATUS "Submodule update")
+        execute_process(COMMAND ${GIT_EXECUTABLE} submodule update --init --recursive
+                        WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}
+                        RESULT_VARIABLE GIT_SUBMOD_RESULT)
+        if(NOT GIT_SUBMOD_RESULT EQUAL "0")
+            message(FATAL_ERROR "git submodule update --init --recursive failed with ${GIT_SUBMOD_RESULT}, please checkout submodules")
+        endif()
+    endif()
+endif()
```
nealkruis: This should probably be replaced with the [newer approach used in CSE](https://github.com/cse-sim/cse/blob/86e65c39c3b2cb4adb9f199aadf02f55c5eb18f2/vendor/CMakeLists.txt#L1-L2).

vendor/CMakeLists.txt
```diff
@@ -0,0 +1,17 @@
...
+
+
+add_subdirectory(btwxt)
```
nealkruis: You need to guard against the possibility that btwxt is already a dependency in the parent program. See this example for [fmt in Kiva](https://github.com/bigladder/kiva/blob/eff0313b15cd48920a6bba404241913493269658/vendor/CMakeLists.txt#L4-L6).

nealkruis (changes requested):

vendor/CMakeLists.txt
pmskintner: Got it. @chipbarnaby  This should prevent HPWHsim from duplicating btwxt builds in CSE.

vendor/CMakeLists.txt
pmskintner: Happy to update it, but just for my clarification, does initialize_submodules just catch the case when submodule directory doesn't yet exist? Which could likely happen for new installs?

nealkruis: Yeah, it just makes sure that the submodules are cloned if they haven't been already.

nealkruis:
@pmskintner I'm going to try to build this in CSE and make sure it builds properly unless you already have.

pmskintner:
> @pmskintner I'm going to try to build this in CSE and make sure it builds properly unless you already have.
@nealkruis  Please do, I haven't done so

nealkruis (approved):
@pmskintner this builds fine with a minor tweak in CSE (we need CSE to build btwxt before HPWHSim tries to).

Other than that, it's just @chipbarnaby's comments regarding how you include Btwxt.

pmskintner:
> @pmskintner this builds fine with a minor tweak in CSE (we need CSE to build btwxt before HPWHSim tries to).
> 
> Other than that, it's just @chipbarnaby's comments regarding how you include Btwxt.

I think that should be covered in [Move btwxt include out of header].(https://github.com/EcotopeResearch/HPWHsim/pull/142/commits/699ff593e82c0f4bd8b27a55ad2a13abbb628e17) I put the include in the source file, and just have a reference just to the RegularGridOperator in the btwxt namespace. If there's something else I'm missing please let me know.

chipbarnaby (approved):

PR #143, opened 2022-04-19
Update CAHP120 UA

pmskintner:


chipbarnaby:
What about incrementing the version #?

Without that change, there is no (simple) way for a user to know whether
the fix is present or not.

pmskintner:
Consider it done

PR #152, opened 2022-12-03
Add overload getSoCFraction()

pmskintner:
- Returns value calculated from logic with the set temperatures.
- Remove unused node weight function


chipbarnaby (commented):
Can't you save the most recently calculated SoC value in the class?  Then the getSoCFraction just returns that value.  Or am I missing something?

Change the primrary fcn name to CalcSocCFraction() perhaps.  Then getSoCFraction() is a simple getter.  Or make a new variant getLatestSoCFraction() that is the simple getter.

pmskintner:
> Can't you save the most recently calculated SoC value in the class? Then the getSoCFraction just returns that value. Or am I missing something?
> 
> Change the primrary fcn name to CalcSocCFraction() perhaps. Then getSoCFraction() is a simple getter. Or make a new variant getLatestSoCFraction() that is the simple getter.

@chipbarnaby, definitely possible to do. I'll just have to watch out for the case where a call to calcSoCFraction(double tMains_C, double tMinUseful_C), does not set the current SoC based off the inputs.

pmskintner:
@chipbarnaby  updated, and cut a bunch of calls calculating the SoC

chipbarnaby (approved):

PR #156, opened 2022-12-11
Clean calls to auto

pmskintner:


chipbarnaby (commented):
I am not a range for expert, but ...
Can't you just use a reference rather than the more complicated shared_ptr?
I believe you can say
for (auto& logic : setOfSources[ i].turnOnLogicSet)

I may be wrong.  But there is no sharing in play here as I see it -- you just need to access an existing object in the data structure.

pmskintner:
The type in turnOnLogicSet is a shared_ptr<HeatingLogic>, I'm not casting to shared_ptr in this case, I'm just explicitly defining the expect type for the loop.

chipbarnaby (approved):

PR #189, opened 2023-12-11
Update HPWHsim to use btwxt v1.0.0

chipbarnaby:
Convert btwxt calls to updated API.

This initial version duplicates prior functionality by updating btwxt calls.  As with the prior version, no error callback scheme is implemented.  An issue has been added to note this.

All HPWHsim regression test pass successfully.


src/CMakeLists.txt
```diff
@@ -26,10 +25,13 @@ set(source
 )
 add_library(libHPWHsim ${source} ${headers})
 
+target_include_directories(libHPWHsim PUBLIC ${PROJECT_BINARY_DIR}/src ${PROJECT_SOURCE_DIR}/src ${PROJECT_SOURCE_DIR}/vendor/btwxt/src)
```
nealkruis: Should not include the btwxt directory. That is added by linking to btwxt through its PUBLIC include directory.

src/HPWHpresets.cc
```diff
@@ -1661,10 +1661,18 @@ int HPWH::HPWHinit_presets(MODELS presetNum) {
...
-		// Set up regular grid interpolator. 
+		// Set up regular grid interpolator.
+#if 0
 		Btwxt::GriddedData gridded_data(compressor.perfGrid, compressor.perfGridValues);
 		gridded_data.set_axis_extrap_method(2, Btwxt::Method::LINEAR); //Linearly extrapolate on Tin (F)
-		compressor.perfRGI = new Btwxt::RegularGridInterpolator(gridded_data);
+#endif
+		Btwxt::GridAxis g0(compressor.perfGrid[0], "TAir", Btwxt::InterpolationMethod::linear, Btwxt::ExtrapolationMethod::linear);
+		Btwxt::GridAxis g1(compressor.perfGrid[1], "TOut", Btwxt::InterpolationMethod::linear, Btwxt::ExtrapolationMethod::linear);
+		Btwxt::GridAxis g2(compressor.perfGrid[2], "TIn", Btwxt::InterpolationMethod::linear, Btwxt::ExtrapolationMethod::linear);
```
nealkruis: Previous version specified linear extrapolation on axis two, but let the others default. @tanaya-mankad is going to verify what those defaults were.

nealkruis (changes requested):

src/HPWHpresets.cc
tanaya-mankad: The GridAxis default used to be CONSTANT extrapolation, so there's definitely a change here.

src/CMakeLists.txt
chipbarnaby: Fixed

src/HPWHpresets.cc
chipbarnaby: Changed TAir and TOut to constant.

nealkruis:
@chipbarnaby should also update the README reference to the Btwxt version (maybe remove the version altogether?).

nealkruis (approved):

PR #205, opened 2024-04-18
Use courier for logging and error handling

spahrenk:
## Description

Enabled logging and error handling using Courier. Removed `verbosity` and `msg` elements and functions. Removed `simHasFailed` member variable and `HPWH_ABORT` constant. Return values are used only for program data; all return codes have been removed and replaced with exceptions or warnings. In the new approach, external calls to most HPWH functions should be enclosed in try/catch.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -41,7 +41,7 @@ endif ()
 
 target_include_directories(${PROJECT_NAME} PUBLIC ${PROJECT_BINARY_DIR}/src ${PROJECT_SOURCE_DIR}/src)
 
-target_link_libraries(${PROJECT_NAME} PRIVATE ${PROJECT_NAME}_common_interface PUBLIC btwxt courier)
+target_link_libraries(${PROJECT_NAME} PRIVATE ${PROJECT_NAME}_common_interface PUBLIC btwxt)
 target_compile_features(${PROJECT_NAME} PRIVATE cxx_std_17)
 
 include(GenerateExportHeader)
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -41,7 +41,7 @@ endif ()
 
 target_include_directories(${PROJECT_NAME} PUBLIC ${PROJECT_BINARY_DIR}/src ${PROJECT_SOURCE_DIR}/src)
 
-target_link_libraries(${PROJECT_NAME} PRIVATE ${PROJECT_NAME}_common_interface PUBLIC btwxt)
+target_link_libraries(${PROJECT_NAME} PRIVATE ${PROJECT_NAME}_common_interface PUBLIC btwxt courier)
 target_compile_features(${PROJECT_NAME} PRIVATE cxx_std_17)
 
 include(GenerateExportHeader)
--- a/src/HPWH.cc
+++ b/src/HPWH.cc
@@ -401,9 +401,8 @@ void HPWH::setMinutesPerStep(const double minutesPerStep_in)
 }
 
 // public HPWH functions
-HPWH::HPWH(const std::shared_ptr<Courier::Courier>& courier) : Sender("name", courier)
+HPWH::HPWH(const std::shared_ptr<Courier::Courier>& courier) : Sender("HPWH", "name", courier)
 {
-    class_name = "HPWH";
     setAllDefaults();
 }
 
@@ -437,7 +436,7 @@ void HPWH::setAllDefaults()
     heatExchangerEffectiveness = 0.9;
 }
 
-HPWH::HPWH(const HPWH& hpwh) : Sender("HPWH", hpwh.courier) { *this = hpwh; }
+HPWH::HPWH(const HPWH& hpwh) : Sender("HPWH", "", hpwh.courier) { *this = hpwh; }
 
 HPWH& HPWH::operator=(const HPWH& hpwh)
 {
--- a/src/HPWH.cc
+++ b/src/HPWH.cc
@@ -375,7 +375,7 @@ void HPWH::setAllDefaults()
     heatExchangerEffectiveness = 0.9;
 }
 
-HPWH::HPWH(const HPWH& hpwh) { *this = hpwh; }
+HPWH::HPWH(const HPWH& hpwh): Sender("HPWH", hpwh.courier) { *this = hpwh; }
 
 HPWH& HPWH::operator=(const HPWH& hpwh)
 {
--- a/src/HPWH.cc
+++ b/src/HPWH.cc
@@ -460,9 +460,9 @@ int HPWH::runOneStep(double drawVolume_L,
     // returns 0 on successful completion, HPWH_ABORT on failure
 
     // check for errors
-    if (doTempDepression == true && minutesPerStep != 1)
+    if (doTempDepression && (minutesPerStep != 1))
     {
-        LOG_WARNING(this, "minutesPerStep must equal one for temperature depression to work.")
+        LOG_ERROR(this, "minutesPerStep must equal one for temperature depression to work.")
         simHasFailed = true;
         return HPWH_ABORT;
     }
@@ -477,7 +477,7 @@ int HPWH::runOneStep(double drawVolume_L,
     // is the failure flag is set, don't run
     if (simHasFailed)
     {
-        LOG_WARNING(this, "simHasFailed is set, aborting.");
+        LOG_ERROR(this, "simHasFailed is set, aborting.");
         return HPWH_ABORT;
     }
 
@@ -803,7 +803,7 @@ int HPWH::runOneStep(double drawVolume_L,
 
     if (simHasFailed)
     {
-        LOG_WARNING(this, "The simulation has encountered an error.");
+        LOG_ERROR(this, "The simulation has encountered an error.");
         return HPWH_ABORT;
     }
 
@@ -840,10 +840,9 @@ int HPWH::runNSteps(int N,
 
         if (simHasFailed)
         {
-            LOG_WARNING(
-                this,
-                "RunNSteps has encountered an error on step {} of N and has ceased running.",
-                i + 1)
+            LOG_ERROR(this,
+                      "RunNSteps has encountered an error on step {} of N and has ceased running.",
+                      i + 1)
             return HPWH_ABORT;
         }
 
@@ -1255,7 +1254,7 @@ int HPWH::setAirFlowFreedom(double fanFraction)
 {
     if (fanFraction < 0 || fanFraction > 1)
     {
-        LOG_WARNING(this, "You have attempted to set the fan fraction outside of bounds.");
+        LOG_ERROR(this, "You have attempted to set the fan fraction outside of bounds.");
         simHasFailed = true;
         return HPWH_ABORT;
     }
@@ -1391,7 +1390,7 @@ int HPWH::setTankSize(double HPWH_size, UNITS units /*=UNITS_L*/, bool forceChan
     }
     if (HPWH_size <= 0)
     {
-        LOG_WARNING(this, "You have attempted to set the tank volume outside of bounds.");
+        LOG_ERROR(this, "You have attempted to set the tank volume outside of bounds.");
         simHasFailed = true;
         return HPWH_ABORT;
     }
@@ -3605,8 +3604,8 @@ bool HPWH::isEnergyBalanced(const double drawVol_L,
     double fracEnergyDiff = fabs(qBal_kJ) / std::max(prevHeatContent_kJ, 1.);
     if (fracEnergyDiff > fracEnergyTolerance)
     {
-        LOG_INFO(this,
-                 fmt::format("Energy-balance error: {} kJ, {} %%", qBal_kJ, 100. * fracEnergyDiff))
+        LOG_WARNING(
+            this, fmt::format("Energy-balance error: {} kJ, {} %%", qBal_kJ, 100. * fracEnergyDiff))
         return false;
     }
     return true;
@@ -4224,7 +4223,7 @@ int HPWH::initFromFile(string configFile)
     inputFILE.open(configFile.c_str());
     if (!inputFILE.is_open())
     {
-        LOG_WARNING(this, "Input file failed to open.")
+        LOG_ERROR(this, "Input file failed to open.")
         return HPWH_ABORT;
     }
 
@@ -4266,7 +4265,7 @@ int HPWH::initFromFile(string configFile)
                 ; // do nothing, lol
             else
             {
-                LOG_WARNING(this, "Incorrect units specification for {}.", token.c_str())
+                LOG_ERROR(this, "Incorrect units specification for {}.", token.c_str())
                 return HPWH_ABORT;
             }
             tankVolume_L = tempDouble;
@@ -4276,7 +4275,7 @@ int HPWH::initFromFile(string configFile)
             line_ss >> tempDouble >> units;
             if (units != "kJperHrC")
             {
-                LOG_WARNING(this, "Incorrect units specification for {}.", token.c_str())
+                LOG_ERROR(this, "Incorrect units specification for {}.", token.c_str())
                 return HPWH_ABORT;
             }
             tankUA_kJperHrC = tempDouble;
@@ -4294,7 +4293,7 @@ int HPWH::initFromFile(string configFile)
             }
             else
             {
-                LOG_WARNING(this, "Improper value for {}", token.c_str())
+                LOG_ERROR(this, "Improper value for {}", token.c_str())
                 return HPWH_ABORT;
             }
         }
@@ -4311,7 +4310,7 @@ int HPWH::initFromFile(string configFile)
             }
             else
             {
-                LOG_WARNING(this, "Improper value for {}.", token.c_str())
+                LOG_ERROR(this, "Improper value for {}.", token.c_str())
                 return HPWH_ABORT;
             }
         }
@@ -4320,7 +4319,7 @@ int HPWH::initFromFile(string configFile)
             line_ss >> tempDouble;
             if (tempDouble < 0 || tempDouble > 1)
             {
-                LOG_WARNING(
+                LOG_ERROR(
                     this, "Out of bounds value for {}. Should be between 0 and 1.", token.c_str())
                 return HPWH_ABORT;
             }
@@ -4335,7 +4334,7 @@ int HPWH::initFromFile(string configFile)
                 ; // do nothing, lol
             else
             {
-                LOG_WARNING(this, "Incorrect units specification for {}.", token.c_str())
+                LOG_ERROR(this, "Incorrect units specification for {}.", token.c_str())
                 return HPWH_ABORT;
             }
             setpoint_C = tempDouble;
@@ -4350,7 +4349,7 @@ int HPWH::initFromFile(string configFile)
                 setpointFixed = false;
             else
             {
-                LOG_WARNING(this, "Improper value for {}", token.c_str())
+                LOG_ERROR(this, "Improper value for {}", token.c_str())
                 return HPWH_ABORT;
             }
         }
@@ -4363,7 +4362,7 @@ int HPWH::initFromFile(string configFile)
                 ;
             else
             {
-                LOG_WARNING(this, "Incorrect units specification for {}.", token.c_str())
+                LOG_ERROR(this, "Incorrect units specification for {}.", token.c_str())
                 return HPWH_ABORT;
             }
             initalTankT_C = tempDouble;
@@ -4379,7 +4378,7 @@ int HPWH::initFromFile(string configFile)
                 hasHeatExchanger = false;
             else
             {
-                LOG_WARNING(this, "Improper value for {}", token.c_str())
+                LOG_ERROR(this, "Improper value for {}", token.c_str())
                 return HPWH_ABORT;
             }
         }
@@ -4410,7 +4409,7 @@ int HPWH::initFromFile(string configFile)
             }
             else
             {
-                LOG_WARNING(this, "Incorrect verbosity on input.")
+                LOG_ERROR(this, "Incorrect verbosity on input.")
                 return HPWH_ABORT;
             }
         }
@@ -4428,7 +4427,7 @@ int HPWH::initFromFile(string configFile)
         {
             if (numHeatSources == 0)
             {
-                LOG_WARNING(
+                LOG_ERROR(
                     this,
                     "You must specify the number of heatsources before setting their properties.")
                 return HPWH_ABORT;
@@ -4443,10 +4442,10 @@ int HPWH::initFromFile(string configFile)
                     heatSources[heatsource].isVIP = false;
                 else
                 {
-                    LOG_WARNING(this,
-                                "Improper value for {} for heat source {}.",
-                                token.c_str(),
-                                heatsource)
+                    LOG_ERROR(this,
+                              "Improper value for {} for heat source {}.",
+                              token.c_str(),
+                              heatsource)
                     return HPWH_ABORT;
                 }
             }
@@ -4459,10 +4458,10 @@ int HPWH::initFromFile(string configFile)
                     heatSources[heatsource].isOn = false;
                 else
                 {
-                    LOG_WARNING(this,
-                                "Improper value for {} for heat source {}.",
-                                token.c_str(),
-                                heatsource)
+                    LOG_ERROR(this,
+                              "Improper value for {} for heat source {}.",
+                              token.c_str(),
+                              heatsource)
                     return HPWH_ABORT;
                 }
             }
@@ -4475,7 +4474,7 @@ int HPWH::initFromFile(string configFile)
                     ; // do nothing, lol
                 else
                 {
-                    LOG_WARNING(this, "Incorrect units specification for {}.", token.c_str())
+                    LOG_ERROR(this, "Incorrect units specification for {}.", token.c_str())
                     return HPWH_ABORT;
                 }
                 heatSources[heatsource].minT = tempDouble;
@@ -4489,7 +4488,7 @@ int HPWH::initFromFile(string configFile)
                     ; // do nothing, lol
                 else
                 {
-                    LOG_WARNING(this, "Incorrect units specification for {}.", token.c_str())
+                    LOG_ERROR(this, "Incorrect units specification for {}.", token.c_str())
                     return HPWH_ABORT;
                 }
                 heatSources[heatsource].maxT = tempDouble;
@@ -4508,12 +4507,11 @@ int HPWH::initFromFile(string configFile)
                         int nodeNum = std::stoi(nextToken);
                         if (nodeNum > LOGIC_SIZE + 1 || nodeNum < 0)
                         {
-                            LOG_WARNING(
-                                this,
-                                "Node number for heatsource {} {} must be between 0 and {}.",
-                                heatsource,
-                                token.c_str(),
-                                LOGIC_SIZE + 1)
+                            LOG_ERROR(this,
+                                      "Node number for heatsource {} {} must be between 0 and {}.",
+                                      heatsource,
+                                      token.c_str(),
+                                      LOGIC_SIZE + 1)
                             return HPWH_ABORT;
                         }
                         nodeNums.push_back(nodeNum);
@@ -4537,7 +4535,7 @@ int HPWH::initFromFile(string configFile)
                     }
                     if (nodeNums.size() != weights.size())
                     {
-                        LOG_WARNING(
+                        LOG_ERROR(
                             this,
                             "Number of weights for heatsource {} {} ({}) does not match number "
                             "of nodes for {} ({}).",
@@ -4550,12 +4548,12 @@ int HPWH::initFromFile(string configFile)
                     }
                     if (nextToken != "absolute" && nextToken != "relative")
                     {
-                        LOG_WARNING(this,
-                                    "Improper definition, \"{}\", for heat source {} {}. Should be "
-                                    "\"relative\" or \"absoute\".",
-                                    nextToken.c_str(),
-                                    heatsource,
-                                    token.c_str())
+                        LOG_ERROR(this,
+                                  "Improper definition, \"{}\", for heat source {} {}. Should be "
+                                  "\"relative\" or \"absoute\".",
+                                  nextToken.c_str(),
+                                  heatsource,
+                                  token.c_str())
                         return HPWH_ABORT;
                     }
                     bool absolute = (nextToken == "absolute");
@@ -4568,12 +4566,12 @@ int HPWH::initFromFile(string configFile)
                         compare = std::greater<double>();
                     else
                     {
-                        LOG_WARNING(this,
-                                    "Improper comparison, \"%s\", for heat source %d %s. Should be "
-                                    "\"<\" or \">\".\n",
-                                    compareStr.c_str(),
-                                    heatsource,
-                                    token.c_str())
+                        LOG_ERROR(this,
+                                  "Improper comparison, \"%s\", for heat source %d %s. Should be "
+                                  "\"<\" or \">\".\n",
+                                  compareStr.c_str(),
+                                  heatsource,
+                                  token.c_str())
                         return HPWH_ABORT;
                     }
                     if (units == "F")
@@ -4591,10 +4589,10 @@ int HPWH::initFromFile(string configFile)
                         ; // do nothing, lol
                     else
                     {
-                        LOG_WARNING(this,
-                                    "Incorrect units specification for {} from heatsource {}.",
-                                    token.c_str(),
-                                    heatsource)
+                        LOG_ERROR(this,
+                                  "Incorrect units specification for {} from heatsource {}.",
+                                  token.c_str(),
+                                  heatsource)
                         return HPWH_ABORT;
                     }
                     std::vector<NodeWeight> nodeWeights;
@@ -4636,7 +4634,7 @@ int HPWH::initFromFile(string configFile)
                             compare = std::greater<double>();
                         else
                         {
-                            LOG_WARNING(
+                            LOG_ERROR(
                                 this,
                                 "Improper comparison, \"%s\", for heat source %d %s. Should be "
                                 "\"<\" or \">\".",
@@ -4667,10 +4665,10 @@ int HPWH::initFromFile(string configFile)
                         ; // do nothing, lol
                     else
                     {
-                        LOG_WARNING(this,
-                                    "Incorrect units specification for %s from heatsource %d.",
-                                    token.c_str(),
-                                    heatsource)
+                        LOG_ERROR(this,
+                                  "Incorrect units specification for %s from heatsource %d.",
+                                  token.c_str(),
+                                  heatsource)
                         return HPWH_ABORT;
                     }
                     if (tempString == "wholeTank")
@@ -4720,7 +4718,7 @@ int HPWH::initFromFile(string configFile)
                     }
                     else
                     {
-                        LOG_WARNING(
+                        LOG_ERROR(
                             this, "Improper %s for heat source %d\n", token.c_str(), heatsource)
                         return HPWH_ABORT;
                     }
@@ -4734,10 +4732,10 @@ int HPWH::initFromFile(string configFile)
                         ; // do nothing, lol
                     else
                     {
-                        LOG_WARNING(this,
-                                    "Incorrect units specification for %s from heatsource %d.",
-                                    token.c_str(),
-                                    heatsource)
+                        LOG_ERROR(this,
+                                  "Incorrect units specification for %s from heatsource %d.",
+                                  token.c_str(),
+                                  heatsource)
                         return HPWH_ABORT;
                     }
                     if (tempString == "topNodeMaxTemp")
@@ -4769,7 +4767,7 @@ int HPWH::initFromFile(string configFile)
                     }
                     else
                     {
-                        LOG_WARNING(
+                        LOG_ERROR(
                             this, "Improper %s for heat source %d\n", token.c_str(), heatsource);
                         return HPWH_ABORT;
                     }
@@ -4788,7 +4786,7 @@ int HPWH::initFromFile(string configFile)
                 }
                 else
                 {
-                    LOG_WARNING(this, "Improper %s for heat source %d\n", token.c_str(), heatsource)
+                    LOG_ERROR(this, "Improper %s for heat source %d\n", token.c_str(), heatsource)
                     return HPWH_ABORT;
                 }
             }
@@ -4809,7 +4807,7 @@ int HPWH::initFromFile(string configFile)
                 }
                 else
                 {
-                    LOG_WARNING(this, "Improper %s for heat source %d", token.c_str(), heatsource)
+                    LOG_ERROR(this, "Improper %s for heat source %d", token.c_str(), heatsource)
                     return HPWH_ABORT;
                 }
             }
@@ -4826,7 +4824,7 @@ int HPWH::initFromFile(string configFile)
                 }
                 else
                 {
-                    LOG_WARNING(this, "Improper %s for heat source %d\n", token.c_str(), heatsource)
+                    LOG_ERROR(this, "Improper %s for heat source %d\n", token.c_str(), heatsource)
                     return HPWH_ABORT;
                 }
             }
@@ -4840,7 +4838,7 @@ int HPWH::initFromFile(string configFile)
                 }
                 else
                 {
-                    LOG_WARNING(this, "Improper %s for heat source %d\n", token.c_str(), heatsource)
+                    LOG_ERROR(this, "Improper %s for heat source %d\n", token.c_str(), heatsource)
                     return HPWH_ABORT;
                 }
             }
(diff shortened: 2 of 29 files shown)


src/CMakeLists.txt
```diff
@@ -41,7 +41,7 @@ endif ()
 
 target_include_directories(${PROJECT_NAME} PUBLIC ${PROJECT_BINARY_DIR}/src ${PROJECT_SOURCE_DIR}/src)
 
-target_link_libraries(${PROJECT_NAME} PRIVATE ${PROJECT_NAME}_common_interface PUBLIC btwxt)
+target_link_libraries(${PROJECT_NAME} PRIVATE ${PROJECT_NAME}_common_interface PUBLIC btwxt courier)
```
nealkruis: This shouldn't be necessary because courier is already part of the btwxt public interface?

src/HPWH.hh
```diff
@@ -25,8 +27,57 @@ class RegularGridInterpolator;
...
+            write_message("DEBUG", message);
+        }
+    }
+    void write_message(const std::string& message_type, const std::string& message)
+    {
+        std::cout << fmt::format("[{}]{} {}", message_type, "", message) << std::endl;
+    }
+};
+
+class HPWH : public Courier::Sender
```
nealkruis: 👍🏻

src/HPWHHeatSources.cc
```diff
@@ -664,32 +572,6 @@ void HPWH::HeatSource::getCapacity(double externalT_C,
...
-
-                if (extrapolate)
-                {
-                    hpwh->msg("Warning performance extrapolation\n\tExternal Temperature: "
-                              "%.2lf\tNearest temperatures:  %.2lf, %.2lf \n\n",
-                              externalT_F,
-                              perfMap[i_prev].T_F,
-                              perfMap[i_next].T_F);
-                }
-            }
```
nealkruis, line 691: Good deletion.

src/HPWH.cc
```diff
@@ -523,43 +507,16 @@ int HPWH::runOneStep(double drawVolume_L,
...
-    if (simHasFailed)
-    {
-        if (hpwhVerbosity >= VRB_reluctant)
-        {
-            msg("simHasFailed is set, aborting.  \n");
-        }
-        return HPWH_ABORT;
+        send_warning(
+            "DR_TOO | DR_TOT use conflicting logic sets. The logic will follow a DR_TOT scheme.");
     }
```
nealkruis, line 520: 👍🏻

src/HPWH.cc
```diff
@@ -523,43 +507,16 @@ int HPWH::runOneStep(double drawVolume_L,
     // returns 0 on successful completion, HPWH_ABORT on failure
 
     // check for errors
-    if (doTempDepression == true && minutesPerStep != 1)
+    if (doTempDepression && (minutesPerStep != 1))
     {
-        msg("minutesPerStep must equal one for temperature depression to work.  \n");
-        simHasFailed = true;
+        send_error("minutesPerStep must equal one for temperature depression to work.");
```
nealkruis, line 519: The return here is not reachable.

src/HPWH.cc
```diff
@@ -786,21 +692,17 @@ int HPWH::runOneStep(double drawVolume_L,
...
-                    // debugging message handling
-                    if (hpwhVerbosity >= VRB_emetic)
-                    {
-                        msg("done heating! runtime_min minutesToRun %.2lf %.2lf\n",
-                            heatSourcePtr->runtime_min,
-                            minutesToRun);
-                    }
+                    send_debug(fmt::format("done heating! runtime_min minutesToRun %.2lf %.2lf\n",
+                                           heatSourcePtr->runtime_min,
+                                           minutesToRun));
```
nealkruis: Can probably be removed entirely.

src/HPWH.cc
```diff
@@ -786,21 +692,17 @@ int HPWH::runOneStep(double drawVolume_L,
...
 
                     // subtract time it ran and turn it off
                     minutesToRun -= heatSourcePtr->runtime_min;
                     heatSources[i].disengageHeatSource();
                     // and if there's a heat source that follows this heat source (regardless of
                     // lockout) that's able to come on,
-                    if (heatSources[i].followedByHeatSource != NULL &&
-                        heatSources[i].followedByHeatSource->shutsOff() == false)
+                    if ((heatSources[i].followedByHeatSource != NULL) &&
+                        !heatSources[i].followedByHeatSource->shutsOff())
```
nealkruis, line 699: 👍🏻

src/HPWH.cc
```diff
@@ -893,10 +795,7 @@ int HPWH::runOneStep(double drawVolume_L,
     // cursory check for inverted temperature profile
     if (tankTemps_C[getNumNodes() - 1] < tankTemps_C[0])
     {
-        if (hpwhVerbosity >= VRB_reluctant)
-        {
-            msg("The top of the tank is cooler than the bottom.  \n");
-        }
+        send_warning("The top of the tank is cooler than the bottom.");
     }
```
nealkruis: I wonder if this entire check can be put into a `#if _DEBUG`?

src/HPWH.cc
```diff
@@ -948,24 +833,22 @@ int HPWH::runNSteps(int N,
...
+                       drawVolume_L[i],
+                       tankAmbientT_C[i],
+                       heatSourceAmbientT_C[i],
+                       DRstatus[i]);
+        }
+        catch (...)
+        {
+            send_warning(fmt::format(
+                "RunNSteps has encountered an error on step {} of N and has ceased running.",
+                i + 1));
```
nealkruis: Not sure if we need to catch this here. If there is an error, probably best for the caller to catch it and decide what to do with it.

src/HPWH.cc
```diff
@@ -732,11 +648,8 @@ int HPWH::runOneStep(double drawVolume_L,
             if (heatSources[i].isLockedOut() && heatSources[i].backupHeatSource == NULL)
             {
                 heatSources[i].disengageHeatSource();
-                if (hpwhVerbosity >= HPWH::VRB_emetic)
-                {
-                    msg("\nWARNING: lock-out triggered, but no backupHeatSource defined. "
-                        "Simulation will continue will lock out the heat source.");
-                }
+                send_warning("lock-out triggered, but no backupHeatSource defined. Simulation will "
+                             "continue will lock out the heat source.");
```
nealkruis: Maybe clean up the message while we're at it? It doesn't read well, and I think there might be a typo in it.

src/HPWH.cc
```diff
@@ -1323,21 +1041,16 @@ int HPWH::setSetpoint(double newSetpoint, UNITS units /*=UNITS_C*/)
     }
     else
     {
-        if (hpwhVerbosity >= VRB_reluctant)
-        {
-            msg("Incorrect unit specification for setSetpoint.  \n");
-        }
+        send_warning(fmt::format("Incorrect unit specification for setSetpoint.", 0));
```
nealkruis: This works for now, but we should probably figure out a way to better handle this with courier. A way to send errors without throwing, so we can accumulate error messages before throwing.

src/HPWH.hh
```diff
@@ -25,8 +27,57 @@ class RegularGridInterpolator;
  * compiled library.  */
 
 #include "HPWHversion.hh"
+#include "courier/helpers.h"
 
-class HPWH
+class Logger : public Courier::Courier
```
nealkruis: I think we should do something very similar to what we are doing in Btwxt for the Default Courier. The idea is that it should be very simple, and it really should be replaced by the caller (e.g., CSE), and the messages should encourage that.

src/HPWHHeatSources.cc
```diff
@@ -896,9 +751,15 @@ void HPWH::HeatSource::regressedMethodMP(double& ynew,
...
-
+    std::vector<double> result;
+    try
+    {
+        result = perfRGI->get_values_at_target(target);
+    }
+    catch (std::string message)
+    {
+        hpwh->send_error(message);
+    }
```
nealkruis: I believe right now it is not possible for Btwxt to throw here because of the way it has been set up. Probably don't need the try/catch.

nealkruis (changes requested):
Looking good! A few changes and discussions needed still, but this is feeling close.

One thing that is missing here is establishing how to set up the courier for the Btwxt objects. I imagine we should use the same courier as the HPWH object that is set by the caller.

Let me know if you'd like to talk through any of the details.

src/HPWH.cc
nealkruis: @spahrenk, do we want to leave this here? I'm not sure it is very helpful.

src/HPWH.cc
```diff
@@ -893,10 +794,7 @@ int HPWH::runOneStep(double drawVolume_L,
     // cursory check for inverted temperature profile
     if (tankTemps_C[getNumNodes() - 1] < tankTemps_C[0])
     {
-        if (hpwhVerbosity >= VRB_reluctant)
-        {
-            msg("The top of the tank is cooler than the bottom.  \n");
-        }
+        send_warning("The top of the tank is cooler than the bottom.");
```
nealkruis: Moving the other comment here. Should this be a debug statement? Under what conditions is this possible? Is it only if inversions are turned off?

src/HPWH.cc
nealkruis: This line also doesn't need `fmt::format`.

src/HPWH.cc
```diff
@@ -1323,21 +1028,16 @@ int HPWH::setSetpoint(double newSetpoint, UNITS units /*=UNITS_C*/)
...
-            msg("Unwilling to set this setpoint for the currently selected model, max setpoint is "
-                "%f C. %s\n",
-                temp,
-                why.c_str());
-        }
+        send_warning(fmt::format(
+            fmt::format("Unwilling to set this setpoint for the currently selected model, "
+                        "max setpoint is {} C. {}",
+                        temp,
+                        why.c_str())));
```
nealkruis: Okay for now, but we should really be formatting the numbers so the messages are cleaner.

src/HPWH.cc
nealkruis: Since these are all returning HPWH_ABORT anyway, maybe they should all turn into `send_error`?

src/HPWH.cc
```diff
@@ -1565,23 +1244,53 @@ double HPWH::getMinOperatingTemp(UNITS units /*=UNITS_C*/) const
...
+/// @param[in]	setTankTemps	new tank temps (arbitrary non-zero size)
+///	@param[in]	units          temp units in setTankTemps (default = UNITS_C)
+/// @return	Success: 0; Failure: HPWH_ABORT
+//-----------------------------------------------------------------------------
+int HPWH::setTankLayerTemperatures(std::vector<double> setTankTemps, const UNITS units)
+{
+    if ((units != UNITS_C) && (units != UNITS_F))
+    {
+        send_warning("Incorrect unit specification for setSetpoint.");
+        return HPWH_ABORT;
```
nealkruis: Probably should just be `send_error`.

src/HPWH.cc
```diff
@@ -1565,23 +1244,53 @@ double HPWH::getMinOperatingTemp(UNITS units /*=UNITS_C*/) const
...
+    }
+
+    // convert setTankTemps to �C, if necessary
+    if (units == UNITS_F)
+        for (auto& T : setTankTemps)
+            T = F_TO_C(T);
+
+    // set node temps
+    if (!resampleIntensive(tankTemps_C, setTankTemps))
+        return HPWH_ABORT;
```
nealkruis: This needs a message and should just be `send_error`.

src/HPWH.cc
```diff
@@ -1663,8 +1369,7 @@ double HPWH::getTankSurfaceArea(UNITS units /*=UNITS_FT2*/) const
     double value = getTankSurfaceArea(tankVolume_L, UNITS_L, units);
     if (value < 0.)
     {
-        if (hpwhVerbosity >= VRB_reluctant)
-            msg("Incorrect unit specification for getTankSurfaceArea.  \n");
+        send_warning(fmt::format("Incorrect unit specification for getTankSurfaceArea.", 0));
```
nealkruis: Nothing to format here.

src/HPWH.cc
```diff
@@ -1716,19 +1420,12 @@ int HPWH::setTankSize(double HPWH_size, UNITS units /*=UNITS_L*/, bool forceChan
...
     }
     if (HPWH_size <= 0)
     {
-        if (hpwhVerbosity >= VRB_reluctant)
-        {
-            msg("You have attempted to set the tank volume outside of bounds.  \n");
-        }
-        simHasFailed = true;
+        send_error("You have attempted to set the tank volume outside of bounds.");
         return HPWH_ABORT;
```
nealkruis: This one is send_error, but also returns HPWH_ABORT, which is not reachable. Please review all these uses to make them consistent (i.e., just `send_error`).

src/HPWH.hh
nealkruis: This still needs to be addressed.

src/HPWHHeatSources.cc
```diff
@@ -60,12 +60,8 @@ HPWH::HeatSource& HPWH::HeatSource::operator=(const HeatSource& hSource)
...
         hSource.followedByHeatSource != NULL)
     {
-        hpwh->simHasFailed = true;
-        if (hpwh->hpwhVerbosity >= VRB_reluctant)
-        {
-            hpwh->msg(
-                "HeatSources cannot be copied if they contain pointers to other HeatSources\n");
-        }
+        hpwh->send_error(
+            "HeatSources cannot be copied if they contain pointers to other HeatSources.");
```
nealkruis: The HPWHHeatSource class should probably be its own Sender class have it's own courier to relay messages up through its parent HPWH.

src/HPWHpresets.cc
```diff
@@ -2201,17 +2153,20 @@ int HPWH::initPreset(MODELS presetNum)
...
-                           "TOut");
+                           "TOut",
+                           get_courier());
         Btwxt::GridAxis g2(compressor.perfGrid[2],
                            Btwxt::InterpolationMethod::linear,
                            Btwxt::ExtrapolationMethod::linear,
                            {-DBL_MAX, DBL_MAX},
-                           "Tin");
+                           "Tin",
+                           get_courier());
```
nealkruis, line 2056: This is good, but the same courier should also be set in the `RegularGridInterpolator` below and anywhere `perfRGI` is instantiated.

nealkruis (changes requested):
This is coming along nicely! There were a few comments that were not addressed, and we need to make sure we are setting all couriers in Btwxt (not just the GridAxes). We also need to review all courier method calls for consistency:

- I don't think there should ever be a `return HPWH_ABORT` after a `send_xxxx` call. They should all just be `send_error`.
- Only use `fmt::format` with strings that need dynamic formatting.
- When using `fmt::format` make sure you are using the correct syntax `{}` vs. `%`. There are several instances where this wasn't addressed and can probably be found with a search of "%".
- Floats should be formatted to be readable (i.e., utilize the decimal formatting features of `fmt`).

spahrenk:
HPWH_ABORT is now removed entirely.

spahrenk:
Please re-review this PR.

src/HPWH.hh
```diff
@@ -25,10 +27,28 @@ class RegularGridInterpolator;
...
+            if (first_occurrence)
+            {
+                std::cout << "  Generated using HPWH::DefaultCourier. Consider deriving your own "
+                             "Courier class!"
+                          << std::endl;
+                first_occurrence = false;
+            }
+        }
+    };
+
```
nealkruis, line 51: What are the implications of nesting this class? In Btwxt it is a separate class, but I'm not sure it matters outside of being consistent. Maybe we should do the same in Btwxt or vice versa?

src/HPWH.cc
```diff
@@ -402,7 +400,7 @@ void HPWH::setMinutesPerStep(const double minutesPerStep_in)
 }
 
 // public HPWH functions
-HPWH::HPWH() : hpwhVerbosity(VRB_silent), messageCallback(NULL), messageCallbackContextPtr(NULL)
+HPWH::HPWH(const std::shared_ptr<Courier::Courier>& courier) : Sender("HPWH", courier)
```
nealkruis: I like this! It is clean and elegant. We should probably do this in Btwxt too.

src/HPWH.cc
```diff
@@ -890,14 +777,13 @@ int HPWH::runOneStep(double drawVolume_L,
             (heatSources[i].energyOutput_kWh - heatSources[i].energyInput_kWh);
     }
 
+#if NDEBUG
```
nealkruis: I like this, but shouldn't this be `#if !NDEBUG`?

This is another area we should probably consider for Courier: Developers probably want to see debug messages when they are testing in release mode. Maybe we need a separate Courier definition (e.g., `COURIER_DEBUG`) that is off by default for release builds, but can be toggled on independent of `NDEBUG`. I'll make a separate issue there. It shouldn't hold this up though.

src/HPWH.cc
```diff
@@ -913,32 +799,16 @@ int HPWH::runOneStep(double drawVolume_L,
...
-        }
-        return HPWH_ABORT;
-    }
-
-    if (hpwhVerbosity >= VRB_typical)
-    {
-        msg("Ending runOneStep.  \n\n\n\n");
-    }
-
     return 0; // successful completion of the step returns 0
```
nealkruis: Should this method also be changed to `void`?

src/HPWH.cc
```diff
@@ -1308,90 +997,70 @@ int HPWH::writeRowAsCSV(std::ofstream& outFILE,
...
-        if (hpwhVerbosity >= VRB_reluctant)
-        {
-            msg("Unwilling to set this setpoint for the currently selected model, max setpoint is "
-                "%f C. %s\n",
-                temp,
-                why.c_str());
-        }
-        return HPWH_ABORT;
+        send_error(fmt::format("Cannot set this setpoint for the currently selected model, "
+                               "max setpoint is {:g} C. {}",
```
nealkruis: We should settle on a consistent format for presenting temperatures. Maybe `{:0.2f}`?

src/HPWH.cc
```diff
@@ -1498,19 +1166,11 @@ double HPWH::calcSoCFraction(double tMains_C, double tMinUseful_C, double tMax_C
     // fractional equation
     if (tMains_C >= tMinUseful_C)
     {
-        if (hpwhVerbosity >= VRB_reluctant)
-        {
-            msg("tMains_C is greater than or equal tMinUseful_C. \n");
-        }
-        return HPWH_ABORT;
+        send_warning("tMains_C is greater than or equal tMinUseful_C.");
```
nealkruis, line 1176: I think it's okay to leave these as they were for now, but we should eventually make them a little more informative (e.g., what are the values of the two temperatures?)

src/HPWH.cc
```diff
@@ -2097,13 +1670,9 @@ int HPWH::setEnteringWaterHighTempShutOff(double highTemp,
...
     {
-        if (hpwhVerbosity >= VRB_reluctant)
-        {
-            msg("High temperature shut off is too close to the setpoint, excpected a minimum "
-                "difference of %.2lf.\n",
-                MINSINGLEPASSLIFT);
-        }
-        return HPWH_ABORT;
+        send_error(fmt::format("High temperature shut off is too close to the setpoint, expected "
+                               "a minimum difference of {:g}",
```
nealkruis, line 1681: Use consistent formatting: `{:0.2f}`.

src/HPWH.cc
```diff
@@ -4694,47 +3916,32 @@ int HPWH::checkInputs()
...
         for (std::shared_ptr<HeatingLogic> logic : heatSources[i].turnOnLogicSet)
         {
             if (!logic->isValid())
             {
-                returnVal = HPWH_ABORT;
-                if (hpwhVerbosity >= VRB_reluctant)
-                {
-                    msg("On logic at index %i is invalid", i);
-                }
+                send_error(fmt::format("On logic at index %i is invalid", i));
```
nealkruis: replace `%i` with `{:d}`

src/HPWH.cc
```diff
@@ -4694,47 +3916,32 @@ int HPWH::checkInputs()
...
         for (std::shared_ptr<HeatingLogic> logic : heatSources[i].shutOffLogicSet)
         {
             if (!logic->isValid())
             {
-                returnVal = HPWH_ABORT;
-                if (hpwhVerbosity >= VRB_reluctant)
-                {
-                    msg("Off logic at index %i is invalid", i);
-                }
+                send_error(fmt::format("Off logic at index %i is invalid", i));
```
nealkruis: replace `%i` with `{:d}`

src/HPWH.cc
```diff
@@ -4790,45 +3984,30 @@ int HPWH::checkInputs()
...
             {
-                if (hpwhVerbosity >= VRB_reluctant)
-                {
-                    msg("Heatsource %d is not an external heat source but has an external "
-                        "secondary heat exchanger. \n",
-                        i);
-                }
-                returnVal = HPWH_ABORT;
+                send_error(
+                    fmt::format("Heatsource %d is not an external heat source but has an external "
```
nealkruis: replace `%d` with `{:d}`

src/HPWH.cc
```diff
@@ -5354,17 +4398,17 @@ int HPWH::initFromFile(string configFile)
...
-                        {
-                            msg("Improper comparison, \"%s\", for heat source %d %s. Should be "
-                                "\"<\" or \">\".\n",
-                                compareStr.c_str(),
-                                heatsource,
-                                token.c_str());
-                        }
-                        return HPWH_ABORT;
+                        send_error(fmt::format(
+                            "Improper comparison, \"%s\", for heat source %d %s. Should be "
```
nealkruis: Need to use fmt syntax, `{}`.

src/HPWH.cc
```diff
@@ -5426,24 +4461,24 @@ int HPWH::initFromFile(string configFile)
...
-                            {
-                                msg("Improper comparison, \"%s\", for heat source %d %s. Should be "
-                                    "\"<\" or \">\".\n",
-                                    compareStr.c_str(),
-                                    heatsource,
-                                    token.c_str());
-                            }
-                            return HPWH_ABORT;
+                            send_error(fmt::format(
+                                "Improper comparison, \"%s\", for heat source %d %s. Should be "
```
nealkruis: Need to use fmt syntax, `{}`.

src/HPWH.cc
```diff
@@ -5513,30 +4539,22 @@ int HPWH::initFromFile(string configFile)
                     }
                     else
                     {
-                        if (hpwhVerbosity >= VRB_reluctant)
-                        {
-                            msg("Improper %s for heat source %d\n", token.c_str(), heatsource);
-                        }
-                        return HPWH_ABORT;
+                        send_error(fmt::format(
+                            "Improper %s for heat source %d\n", token.c_str(), heatsource));
```
nealkruis: Need to use fmt syntax, `{}`.

src/HPWH.cc
```diff
@@ -5566,11 +4584,8 @@ int HPWH::initFromFile(string configFile)
...
```
nealkruis: Need to use fmt syntax, `{}`.

src/HPWH.cc
```diff
@@ -5587,11 +4602,8 @@ int HPWH::initFromFile(string configFile)
                 }
                 else
                 {
-                    if (hpwhVerbosity >= VRB_reluctant)
-                    {
-                        msg("Improper %s for heat source %d\n", token.c_str(), heatsource);
-                    }
-                    return HPWH_ABORT;
+                    send_error(
+                        fmt::format("Improper %s for heat source %d\n", token.c_str(), heatsource));
```
nealkruis: Need to use fmt syntax, `{}`.

src/HPWH.cc
```diff
@@ -5611,11 +4623,8 @@ int HPWH::initFromFile(string configFile)
                 }
                 else
                 {
-                    if (hpwhVerbosity >= VRB_reluctant)
-                    {
-                        msg("Improper %s for heat source %d\n", token.c_str(), heatsource);
-                    }
-                    return HPWH_ABORT;
+                    send_error(
+                        fmt::format("Improper %s for heat source %d", token.c_str(), heatsource));
```
nealkruis: Need to use fmt syntax, `{}`.

src/HPWH.cc
```diff
@@ -5631,11 +4640,8 @@ int HPWH::initFromFile(string configFile)
                 }
                 else
                 {
-                    if (hpwhVerbosity >= VRB_reluctant)
-                    {
-                        msg("Improper %s for heat source %d\n", token.c_str(), heatsource);
-                    }
-                    return HPWH_ABORT;
+                    send_error(
+                        fmt::format("Improper %s for heat source %d\n", token.c_str(), heatsource));
```
nealkruis: Need to use fmt syntax, `{}`.

src/HPWH.cc
```diff
@@ -5648,11 +4654,8 @@ int HPWH::initFromFile(string configFile)
...
```
nealkruis: Need to use fmt syntax, `{}`.

src/HPWH.cc
```diff
@@ -5664,14 +4667,10 @@ int HPWH::initFromFile(string configFile)
...
```
nealkruis: Need to use fmt syntax, `{}`.

src/HPWH.cc
```diff
@@ -5696,45 +4695,33 @@ int HPWH::initFromFile(string configFile)
...
                     {
-                        if (hpwhVerbosity >= VRB_reluctant)
-                        {
-                            msg("%s specified for heatsource %d before definition of nTemps.  \n",
-                                token.c_str(),
-                                heatsource);
-                        }
-                        return HPWH_ABORT;
+                        send_error(fmt::format(
+                            "%s specified for heatsource %d before definition of nTemps.",
```
nealkruis: Need to use fmt syntax, `{}`.

src/HPWH.cc
```diff
@@ -5696,45 +4695,33 @@ int HPWH::initFromFile(string configFile)
...
-                                "less than %d.  \n",
-                                token.c_str(),
-                                heatsource,
-                                maxTemps,
-                                nTemps);
-                        }
-                        return HPWH_ABORT;
+                        send_error(fmt::format(
+                            "Incorrect specification for %s from heatsource %d. nTemps, %d, is "
+                            "less than %d.  \n",
```
nealkruis: Need to use fmt syntax, `{}`.

src/HPWH.cc
```diff
@@ -5768,26 +4755,20 @@ int HPWH::initFromFile(string configFile)
...
                     {
-                        if (hpwhVerbosity >= VRB_reluctant)
-                        {
-                            msg("%s specified for heatsource %d before definition of nTemps.  \n",
-                                token.c_str(),
-                                heatsource);
-                        }
-                        return HPWH_ABORT;
+                        send_error(fmt::format(
+                            "%s specified for heatsource %d before definition of nTemps.",
```
nealkruis: Need to use fmt syntax, `{}`.

src/HPWH.cc
```diff
@@ -5768,26 +4755,20 @@ int HPWH::initFromFile(string configFile)
...
-                                "less than %d.  \n",
-                                token.c_str(),
-                                heatsource,
-                                maxTemps,
-                                nTemps);
-                        }
-                        return HPWH_ABORT;
+                        send_error(fmt::format(
+                            "Incorrect specification for %s from heatsource %d. nTemps, %d, is "
+                            "less than %d.  \n",
```
nealkruis: Need to use fmt syntax, `{}`.

src/HPWH.cc
```diff
@@ -5838,17 +4814,14 @@ int HPWH::initFromFile(string configFile)
             }
             else
             {
-                if (hpwhVerbosity >= VRB_reluctant)
-                {
-                    msg("Improper specifier (%s) for heat source %d\n", token.c_str(), heatsource);
-                }
+                send_error(fmt::format(
+                    "Improper specifier (%s) for heat source %d\n", token.c_str(), heatsource));
```
nealkruis: Need to use fmt syntax, `{}`.

src/HPWH.cc
```diff
@@ -5838,17 +4814,14 @@ int HPWH::initFromFile(string configFile)
...
             }
 
         } // end heatsource options
         else
         {
-            msg("Improper keyword: %s  \n", token.c_str());
-            return HPWH_ABORT;
+            send_error(fmt::format("Improper keyword: %s", token.c_str()));
```
nealkruis: Need to use fmt syntax, `{}`.

src/HPWH.cc
```diff
@@ -4277,10 +3505,8 @@ bool HPWH::isEnergyBalanced(const double drawVol_L,
     double fracEnergyDiff = fabs(qBal_kJ) / std::max(prevHeatContent_kJ, 1.);
     if (fracEnergyDiff > fracEnergyTolerance)
     {
-        if (hpwhVerbosity >= VRB_reluctant)
-        {
-            msg("Energy-balance error: %f kJ, %f %% \n", qBal_kJ, 100. * fracEnergyDiff);
-        }
+        send_warning(
+            fmt::format("Energy-balance error: {:g} kJ, {:g} %%", qBal_kJ, 100. * fracEnergyDiff));
```
nealkruis: I imagine the first `%` is an escape for the old format specification?

src/HPWHHeatSources.cc
nealkruis: This message was marked resolved, but I don't see any changes.

src/HPWHHeatSources.cc
```diff
@@ -160,68 +158,32 @@ bool HPWH::HeatSource::shouldLockOut(double heatSourceAmbientT_C) const
...
-        //		if (hpwh->hpwhVerbosity >= HPWH::VRB_emetic) {
-        //			hpwh->msg("\nWARNING: lock-out triggered, but no backupHeatSource defined.
-        // Simulation will continue without lock-out");
-        //		}
-        //		lock = false;
-        //	}
-        if (hpwh->hpwhVerbosity >= VRB_typical)
-        {
-            hpwh->msg("\n");
+            hpwh->send_warning(fmt::format("lock-out: condenser water temperature above max: {:g}",
```
nealkruis: Add consistent precision?

src/HPWHHeatSources.cc
```diff
@@ -529,11 +400,12 @@ void HPWH::HeatSource::addHeat(double externalT_C, double minutesToRun)
...
         runtime_min = (1. - (leftoverCap_kJ / cap_kJ)) * minutesToRun;
-#if 1 // error check, 1-22-2017; updated 12-6-2023
+
         if (runtime_min < -TOL_MINVALUE)
-            if (hpwh->hpwhVerbosity >= VRB_reluctant)
-                hpwh->msg("Internal error: Negative runtime = %0.3f min\n", runtime_min);
-#endif
+        {
+            hpwh->send_error(
+                fmt::format("Internal error: Negative runtime = {:g} min", runtime_min));
```
nealkruis: Make formatting consistent.

src/HPWHHeatSources.cc
```diff
@@ -896,9 +721,8 @@ void HPWH::HeatSource::regressedMethodMP(double& ynew,
 
 void HPWH::HeatSource::btwxtInterp(double& input_BTUperHr, double& cop, std::vector<double>& target)
 {
-
-    std::vector<double> result = perfRGI->get_values_at_target(target);
-
+    std::vector<double> result;
+    result = perfRGI->get_values_at_target(target);
```
nealkruis: Harmless, but curious change...

test/main.cc
```diff
@@ -415,15 +415,22 @@ int main(int argc, char* argv[])
...
+                            drStatus, // DDR Status (now an enum. Fixed for now as allow)
+                            1. * GAL_TO_L(allSchedules[1][i]),
+                            allSchedules[0][i],
+                            vectptr);
+        }
+        catch (...)
+        {
+            cout << "Error in hpwh.runOneStep.\n";
+            exit(1);
+        }
```
nealkruis: Is this necessary? Won't an uncaught exception effectively terminate the program with a non-zero return code? It probably doesn't really matter one way or the other, but it is more code than was there before.

test/unit_tests/compressorFncsTest.cpp
```diff
@@ -47,17 +47,28 @@ TEST_F(CompressorFncsTest, compressorSpecs)
     {
         // get preset model
         HPWH hpwh;
-        EXPECT_EQ(hpwh.initPreset(modelSpec.sModelName), 0)
+        EXPECT_NO_THROW(hpwh.initPreset(modelSpec.sModelName))
```
nealkruis: We can probably just remove the check here (and in most places where we're using `EXPECT_NO_THROW` where we were checking for zero return codes. If the method throws, the test will fail.

nealkruis (changes requested):
Getting closer! I feel the code is much cleaner without checking return codes all over the place!

src/HPWH.cc
nealkruis: I think this question was overlooked?

src/HPWH.cc
nealkruis: Maybe overlooked? For most cases where we are formatting doubles, we should have something rounded so it is easier for users to read.

src/HPWH.cc
nealkruis: Missed this one.

src/HPWH.cc
nealkruis: Missed this one.

src/HPWH.hh
nealkruis: Question still remains.

src/HPWHHeatSources.cc
```diff
@@ -11,8 +11,11 @@
 #include "HPWH.hh"
 
 // public HPWH::HeatSource functions
-HPWH::HeatSource::HeatSource(HPWH* parentInput)
-    : hpwh(parentInput)
+HPWH::HeatSource::HeatSource(
+    HPWH* parentInput,
+    const std::shared_ptr<Courier::Courier> courier /*std::make_shared<DefaultCourier>()*/)
+    : Sender("HeatSource", courier)
+    , hpwh(parentInput)
```
nealkruis: The Sender class also has a `parent_pointer` member that allows the message context to be passed up through the hierarchy of classes. That should be set too.

src/HPWHpresets.cc
```diff
@@ -68,7 +51,7 @@ int HPWH::initResistanceTank(double tankVol_L,
     doTempDepression = false;
     tankMixesOnDraw = true;
 
-    HeatSource resistiveElementBottom(this);
+    HeatSource resistiveElementBottom(this, get_courier());
```
nealkruis: I wonder if we can make a method on HPWH to "add_heat_source" that takes care of assigning the pointer and the courier. It could return a pointer to the heat source so it can be further modified instead of creating the whole thing and copying it into the HPWH heatSources vector?

src/HPWH.cc
nealkruis: Now that I'm looking at this closer, I'm realizing that the constructor needs the instance `name`, and not the `class_name`. The constructor of a Sender class needs the user to define the instance `name`, and needs to set the `class_name` in the body of the constructor.

nealkruis (changes requested):
Getting closer! I'm realizing that the `Sender` class in courier isn't well documented. Part of the idea is that messages that work their way up all attach a bit of their context to the final message. For example: a warning from a grid axis in Btwxt should show up as something like:

`[WARNING] HPWH 'AOSmith': HeatSource 'compressor': RegularGridInterpolator: 'Compressor Map': GridAxis 'TAir': blah blah blah`.

As it stands, this will come out as:

`[WARNING]  'HPWH':  'HeatSource': RegularGridInterpolator: 'RegularGridInterpolator': GridAxis 'TAir': blah blah blah`.

There are a few other unresolved comments/questions to address too.

src/HPWHHeatSources.cc
```diff
@@ -38,15 +40,15 @@ HPWH::HeatSource::HeatSource(HPWH* parentInput, const std::shared_ptr<Courier::C
...
+HPWH::HeatSource::HeatSource(const HeatSource& hSource) : Sender(hSource) { *this = hSource; }
 
 HPWH::HeatSource& HPWH::HeatSource::operator=(const HeatSource& hSource)
 {
     if (this == &hSource)
     {
         return *this;
     }
-    Sender::operator = (hSource);
+    Sender::operator=(hSource);
```
nealkruis, line 53: TODO: Check if this is necessary.

src/HPWH.cc
```diff
@@ -444,7 +444,7 @@ HPWH& HPWH::operator=(const HPWH& hpwh)
         return *this;
     }
 
-    Sender::operator = (hpwh);
+    Sender::operator=(hpwh);
```
nealkruis, line 449: TODO: Check if this is necessary.

src/HPWHHeatSources.cc
spahrenk: This seems to be needed to copy the courier from hSource. (Could use `courier = hSource.courier`..., instead.)

spahrenk: Needed because the HeatSource copy constructor is used to assign configured heat sources from local variables into the heatSources vector member of HPWH. Sender data is not transferred implicitly.

src/HPWH.cc
spahrenk: Seems to be needed if assignment of HPWH objects is performed. However, unlike HeatSources, HPWHs are instantiated with a default courier, so no fault would result without this change.

src/HPWHpresets.cc
spahrenk: Added function HPWH::makeHeatSource.

src/HPWH.hh
```diff
@@ -64,12 +84,19 @@ class HPWH
     static const double
         MINSINGLEPASSLIFT; /**< The minimum temperature lift for single pass compressors */
 
-    HPWH();                            /**< default constructor */
-    HPWH(const HPWH& hpwh);            /**< copy constructor  */
-    HPWH& operator=(const HPWH& hpwh); /**< assignment operator  */
+    HPWH(const std::shared_ptr<Courier::Courier>& courier =
```
nealkruis: The constructor should include an argument for the caller to set the `name` of the object.

src/HPWH.cc
nealkruis: This still needs to be addressed.

src/HPWHHeatSources.cc
```diff
@@ -11,8 +11,12 @@
...
 
 // public HPWH::HeatSource functions
-HPWH::HeatSource::HeatSource(HPWH* parentInput)
-    : hpwh(parentInput)
+HPWH::HeatSource::HeatSource(
+    const std::string& name_in,
+    HPWH* hpwh_in,
+    const std::shared_ptr<Courier::Courier> courier_in /*std::make_shared<DefaultCourier>()*/)
+    : Sender("HeatSource", name_in, courier_in)
+    , hpwh(hpwh_in)
```
nealkruis, line 19: We probably don't need both `hpwh` and `parent_pointer`? What do you think?

src/HPWH.cc
```diff
@@ -512,54 +497,24 @@ HPWH& HPWH::operator=(const HPWH& hpwh)
...
-
-    // check for errors
-    if (doTempDepression == true && minutesPerStep != 1)
-    {
-        msg("minutesPerStep must equal one for temperature depression to work.  \n");
-        simHasFailed = true;
-        return HPWH_ABORT;
-    }
+    return HeatSource(name_in, this, get_courier());
+}
```
nealkruis: What if we made it this instead:

```
HPWH::HeatSource* HPWH::addHeatSource(const std::string& name_in)
{
    heatSources.emplace_back(HeatSource(name_in, this, get_courier());
    return &heatSources.back();
}
```

I think you mentioned some issue, but I wonder if we can support it by adding a little more functionality? This way less copying is required.

src/HPWH.cc
nealkruis: Makes sense. Thanks!

src/HPWHHeatSources.cc
nealkruis: Makes sense! I think Sender data is transferred implicitly in the implicit assignment operator, but since this is an overriding assignment operator, we need to explicitly call the base class operator function.

nealkruis (changes requested):
Just a few more changes/questions!

src/HPWHHeatSources.cc
spahrenk: I don't know what you mean. Don't assign parent_pointer?

spahrenk: I would choose to keep them separate. The two happen to be the same pointer in this case, but the uses and purposes are unrelated, and could evolve. The change would require something like `static_cast<HPWH*>(parent_pointer)`, where we currently have just  `hpwh`. It is not uncommon to retain two pointers that correspond to the same object.

nealkruis (approved):
This is looking good enough to me, but I'm going to hold off on merging into "main" until @chipbarnaby has had a chance to review/modify the integration on the CSE side. I'm wondering if we should start a "develop" branch where we can merge these branches while we're waiting for integration/testing at the CSE level.

PR #209, opened 2024-08-19
Add AeroTherm2023 models

spahrenk:
Bradford White Aerotherm2023 50-, 65-, and 80-gal models are incorporated into HPWHsim. The 50-gal model was refined to match reasonably the RE2H50_UEF50 and RE2H50_UEF67 test schedules extracted from the provided measured datasets. These schedules were translated into the standard HPWHsim, multifile, schedule format, using scripts added to scripts/conversion.

Additional python scripts and an html file were added to facilitate comparison of simulated and measured profiles.  The comparison is done by 1) running `poetry run python server.py` from within the scripts/python folder and 2) opening the file `index.html` from a browser (e.g., Chrome). The user can interactively specify the HPWH model and specification, then apply test parameters or view the 24-hr test output. Presently, a measured test data file (for comparison to simulated test data) must be specified to generate plots.


nealkruis:
@spahrenk is this ready to review again?

spahrenk:
Should be able to change the build folder now.

src/HPWH.hh
```diff
@@ -282,7 +282,10 @@ class HPWH : public Courier::Sender
 
         MODELS_AquaThermAire = 400, // heat exchanger model
 
-        MODELS_GenericUEF217 = 410
+        MODELS_GenericUEF217 = 410,
```
nealkruis: This should probably be added to the other generic models (e.g., number 179).

src/HPWH.hh
```diff
@@ -282,7 +282,10 @@ class HPWH : public Courier::Sender
...
+        MODELS_AeroTherm2023_50 = 411,
+        MODELS_AeroTherm2023_65 = 412,
+        MODELS_AeroTherm2023_80 = 413
```
nealkruis: Start a new series (500?) for Bradford-White AeroTherm. Follow same naming convention as other enumerations:

`MODELS_BradfordWhiteAeroThermRE2H50`

nealkruis (changes requested):
Looking good! The python server is working for me now.

I have a few minor comments, and would like to see the python modules documented with doc strings explaining what the module does and how/when to use them.

src/HPWH.hh
spahrenk: Done.

src/HPWH.hh
spahrenk: Done.

spahrenk: Done.

spahrenk:
Ready to review.

scripts/python/conversion/convert_measurements.py
nealkruis, line 1: I'm still a bit confused by this file. It seems hard-coded to villara. Do we anticipate needing this file for future models? We talked with Ben about setting a standardized template for manufacturers to supply test data. Would that obsolete the need for this file?

scripts/python/conversion/convert_test_schedule.py
nealkruis, line 1: Similar with this file. Is there an example you can point to where this has been used?

scripts/python/conversion/convert_measurements.py
spahrenk: Two formats encountered to date are both implemented. No information on other formats may emerge.

scripts/python/conversion/convert_test_schedule.py
spahrenk: Used for all Aerotherm2023 models provided. Not intended as an automated tool for arbitrary format.

nealkruis (approved):

PR #210, opened 2024-08-23
Improve units handling (rev)

spahrenk:
## Description
- Changes unit-handling throughout HPWHsim.  Independent physical constants relevant to HPWHsim are reduced to a short list of exact or reliable values, from which others are then derived. These are defined in file `units.hpp`, which provides inline initialization of conversion maps and defines partially specialized types for containing values and vectors with unit information. 
- The conversion mechanism is established in 'unity.hpp'. A static std::map is retained for each type of units that provides conversion to/from a reference unit. No association between units of different types (e.g., 'm/s' = 'm' / 's') was deemed necessary to improve units organization within HPWHsim, while maintaining performance.
- Values with units are specializations of `ScaleVal` and `ScaleOffsetVal`, and are retained as doubles. Effort was made to enable easy accessibility of the underlying double, while retaining explicit construction. The internal value (double), or a reference to this value, can be accessed with the () operator.
- The same guidelines were applied to vectors. These are retained internally as `std::vector<double>` (e.g., rather than `std::vector<ScaleVal<...>>`). 
- Every model test reference was replaced, due to generally negligible changes. Differences in yearly test results are somewhat larger, but appear to be within normal tolerances (~ 0.1%). An additional unit test for units conversions is added.


nealkruis:
@spahrenk make new PR merging into main.

PR #211, opened 2024-09-30
Initialize HPWH objects from schema-based data model

spahrenk:
Enables initialization of HPWH objects from IHPWH data-model specification in JSON format. The IHPWH data-model specification is defined by schema entries in repository [hpwh-data-model](https://github.com/bigladder/hpwh-data-model). The schema are used by lattice to generate a code library for reading the JSON representations of the HPWH models. The code-generation step could be initiated by cmake when building the project (rather than being retaining the code in [HPWHsim](https://github.com/bigladder/HPWHsim), but this has not been implemented, due to a relatively small number of manual changes that remain necessary in the generated code.

The command-line utility `convert` was added to the hpwh executable. This was used to convert all existing IHPWH models from preset entries to JSON files. These models were exported to the folder "models_json " in the "test" folder.  Many of these models were previously implemented using polynomial expansions at specific environment temperatures of the variation of input power and cop with heat-source temperature (determined by the condensity-weighted tank temperature.) These were converted by sampling into performance grids for evaluation using `btwxt`. Criteria were developed for the number of grid points, based on the curvatures of the input-power and cop functions, to ensure sufficient resolution of the grid representations for the reproduction of reference test results.

The assigned JSON file names correspond to their predefined model names. However, 'cse' primarily specifies HPWH presets by HPWH::MODELS number, whereas HPWHsim uses only specification by model name, not number, for testing. Furthermore, roughly eight models did not have assigned names, and were inaccessible via command line. To export these models to JSON, the `convert` utility also allows specification by either model name or number.  Obvious names were also defined in the code for these models. However, no mechanism exists currently to initiate a model from a JSON representation via model number.

Tests were added for comparison of the JSON and Preset (and File) representations.


nealkruis:
@spahrenk make new PR merging `use-data-models` into `main`.

PR #215, opened 2024-10-28
Small changes to address review comments

spahrenk:
Contains additional changes from reviews of courier and CLI11 PRs.

### DIFF
--- a/test/unit_tests/sizingFractionsTest.cpp
+++ b/test/unit_tests/sizingFractionsTest.cpp
@@ -22,7 +22,7 @@ TEST(SizingFractionsTest, TamScalable_SP_SizingFract)
     double aquaFrac, useableFrac;
     const double aquaFrac_answer = 4. / logicSize;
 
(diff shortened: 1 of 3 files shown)


test/unit_tests/sizingFractionsTest.cpp
```diff
@@ -22,7 +22,7 @@ TEST(SizingFractionsTest, TamScalable_SP_SizingFract)
     double aquaFrac, useableFrac;
     const double aquaFrac_answer = 4. / logicSize;
 
-    EXPECT_EQ(hpwh.getSizingFractions(aquaFrac, useableFrac), 0);
+    EXPECT_NO_THROW(hpwh.getSizingFractions(aquaFrac, useableFrac));
```
nealkruis: `EXPECT_NO_THROW` isn't necessary since an exception would also cause the test to fail.

nealkruis (changes requested):

nealkruis (approved):

PR #220, opened 2024-12-10
Constrain performance curves and add 80-gallon LG model

spahrenk:
Performance-map polynomials of recently added Bradford White and LG models were adjusted to prevent the use of unreasonable values within the ranges of allowed operating temperatures.
An additional 80-gal LG model has also been added.


nealkruis (approved):

nealkruis (changes requested):
@spahrenk sorry--I missed the fact that the CI isn't happy here. Can you fix the formatting? We likely need to update the MacOS versions we are testing on too.

spahrenk:
Should be formatted now.

nealkruis (approved):

PR #225, opened 2025-01-22
Use new data model for external files

### LINKED ISSUE #228: Generate data-model code in vendor using repo in build
Expand python scripts for building hpwh_data_model using lattice so that

- the generated code is in vendor/hpwh_data_model
- the hpwh_data_model schema repo is in the build directory

Code generation could be handled separately or as part of CMake build.

spahrenk:
NOTE: This PR also includes the changes originally proposed in #214 and #223.

This branch incorporates weighted distributions of the form
``{{height0, height1, ..., height_final},{weight0, weight1, ..., weight_final}}``
into HPWHsim. These replace std::vector arrays for expressing heating-power distributions and logic nodes. (The tank temperature profile remains expressed by a std::vector representing a fixed number of tank node temperatures.) The notation for weighted distributions is generally more flexible and compact than the previous forms, and is amenable to JSON data-model format. Normalization is automatic. The height range is taken as [0., height_final]. Heights are scaled upon export so that height_final = 1. Weights are scaled upon export to give a maximum weight of unity.

Existing distributions in Preset and File models are immediately converted to weighted form during loading. Most internal calculations for applying heat or checking logic temperatures are now performed using weighted logic. The transition is mostly obvious and will produce the same results for existing models. But additional conditions are enforced by the internal construction of these models that are not enforced directly when loading from a JSON data model. The distribution is assumed to be peaked about a certain heating or logic node in certain cases. Specifying a different distribution may produce unpredictable behavior:
1. The function `TempBasedHeatingLogic::getFractToMeetComparisonExternal` is used for single-pass external condensers to prevent exceeding the shut-off condition during heating, but assumes the heating logic refers to a single node out of a fixed 12-node logic structure, which is enforced during initialization of these models. Specifying a different distribution will produce unpredictable behavior.
2. Resistance heat-source elements are organized by position along the tank height, so that particular activation conditions can be applied the lowest element. The lowest non-zero height in the weighted distribution is used in this case. Resistance positions are also used in `ResistanceFunctionsTest`, but no cases were found that affect simulations.


nealkruis (approved):

nealkruis:
@spahrenk any idea why the CI is failing on Mac?

spahrenk:
It seems to try generating the code before it has finished downloading the hpwh_data_model repo. It is usually successful on a second try. I might be able to add a delay or condition to avoid this.

nealkruis:
> It seems to try generating the code before it has finished downloading the hpwh_data_model repo. It is usually successful on a second try. I might be able to add a delay or condition to avoid this.

@spahrenk This should be handled via file dependencies in CMake, and not a manual delay.

spahrenk:
@nealkruis The code generation in the CI seems to be fixed now.

spahrenk:
The generated code is added to the repo in vendor/hpwh_data_model. The code can be re-generated using 'build_data_model.py'. Note that this script refers to a specific branch name or sha of the hpwh-data-model repo for generation, rather than the default (`main`).

vendor/CMakeLists.txt
```diff
@@ -19,3 +19,6 @@ add_submodule(CLI11)
 
 # JSON library
 add_submodule(json)
+
+# hpwh_data_model generated code
+add_subdirectory(hpwh_data_model/cpp)
```
nealkruis, line 24: Why the extra "cpp" directory?

test/models_json/AOSmithCAHP120.json
```diff
@@ -3,30 +3,38 @@
   "fixed_volume": true,
   "integrated_system": {
     "metadata": {
-      "schema": "RSINTEGRATEDWATERHEATER"
+      "author": "auto-generated by HPWHsim",
```
nealkruis: Let's go with "California Energy Commission" for now. We should probably add "Generated by HPWHsim" to a "source" data element.

test/models_json/AOSmithCAHP120.json
```diff
@@ -3,30 +3,38 @@
...
+      "description": "representation of embedded HPWH element",
+      "schema_author": "none",
```
nealkruis: Let's use "Big Ladder Software, LLC" for now. We hope it will eventually be adopted by ASHRAE. We also should add the "schema_url" data element.

test/models_json/AOSmithCAHP120.json
```diff
@@ -3,30 +3,38 @@
   "fixed_volume": true,
   "integrated_system": {
     "metadata": {
-      "schema": "RSINTEGRATEDWATERHEATER"
+      "author": "auto-generated by HPWHsim",
+      "description": "representation of embedded HPWH element",
```
nealkruis: This is not an accurate description for an Integrated Heat Pump Water Heater. This should be a unique description for each representation.

test/models_json/AOSmithCAHP120.json
```diff
@@ -3,30 +3,38 @@
...
+      "schema_author": "none",
+      "schema_name": "RSINTEGRATEDWATERHEATER",
+      "schema_version": "0.1.0",
+      "time_of_creation": "UTC: 22:08:11 2025"
```
nealkruis: This should follow the pattern defined for [Timestamp](https://github.com/bigladder/lattice/blob/5412f7135e3fdc6a14dafaa0c8b73bb6a2e750f3/lattice/core.schema.yaml#L70). Not sure how this file is valid. @tanaya-mankad, can you look into this?

nealkruis (changes requested):
Pretty minor changes requested here. We'll need to figure out how to flesh out the descriptions of each representation eventually along with populating the optional product information--and eventually ratings, which we haven't developed the schema for yet.

vendor/CMakeLists.txt
spahrenk: That is generated by lattice, along with the empty schema folders. Should this be reorganized after generation?

nealkruis: @tanaya-mankad, doesn't lattice have the ability to specify a directory for the CPP generated code? I think by default it should appear in .lattice/cpp, but the user should be able to specify a different directory.

tanaya-mankad: It's not implemented yet. I can make that my next priority.

test/models_json/AOSmithCAHP120.json
spahrenk: Link to the original yaml files?

vendor/CMakeLists.txt
nealkruis: Please do, so we can get this PR cleaned up. We might want output directory overrides for other Lattice outputs too?

test/models_json/AOSmithCAHP120.json
spahrenk: No, it is duplicated from the schema. There are no descriptions available, to my knowledge, other than scattered notes. Do you want me to write the descriptions? The description could be retained in a HPWH member variable.

test/models_json/AOSmithCAHP120.json
nealkruis: Yes. e.g.: https://github.com/bigladder/hpwh-data-model/blob/main/schema/RSINTEGRATEDWATERHEATER.schema.yaml

test/models_json/AOSmithCAHP120.json
nealkruis: I think the description should ultimately be generated from other data elements (e.g., Manufacturer, model number, nominal tank volume, UEF). These should also be used as the hash seed to generate the UUID for the "id". However, I'm not sure all of that needs to be managed by the C++ code, though. I imagine some of this could be handled by python after HPWHsim generates the performance data.

Let's do something fairly basic for now, but before this all merges in it would be good to have this process planned out.

test/models_json/AOSmithCAHP120.json
tanaya-mankad: It doesn't validate. But we don't have validation in the C++ code (it's an issue that we put on the back burner a while ago) so it's not being checked there.
For the record, this file also shows 2 more errors with lattice:
  1. 'central_system' is a required property ()
  2. Additional properties are not allowed ('metadata' was unexpected) ()

nealkruis: It should be caught in [hpwh-data-model](https://github.com/bigladder/hpwh-data-model/pull/15).

vendor/CMakeLists.txt
tanaya-mankad: Does it work to add a Path parameter to the Lattice constructor? As we have no other getters or setters, this seems clean. @spahrenk, do you know the cpp directory you want before your Lattice object is created?

nealkruis: @tanaya-mankad can it just be an argument to [generate_cpp_project](https://github.com/bigladder/lattice/blob/9847e6d130ff694c18298fc0f05d3637edc8778a/lattice/lattice.py#L260)?

spahrenk: The generated code could be placed directly into vendor/hpwh_data_model. On the other hand, it might also be useful to populate the schema folders that are currently created there.

nealkruis:
@spahrenk I think this is just about ready. Can you merge main in here to bring it up to date and resolve any resulting conflicts?

nealkruis (approved):

spahrenk:
@nealkruis This should be ready to merge into main.

nealkruis:
Thanks, @spahrenk! We still want to merge into `add_cwhs`, but we needed `main` merged into this first.

PR #226, opened 2025-02-26
Large draw fix

spahrenk:
Errors were reported in cse using draw volumes exceeding the tank volume. This executes a section of code in HPWH::update_tank_temps for handling evolution of the tank-node temperatures during the draw. This large-draw section was intended to be carried on, but is not executed in any tests. (A 1-min draw exceeding the tank volume is uncommon; a subdivision of the draw may offer a more realistic simulation.) A bug may have been introduced within this section during [balance-energy](https://github.com/bigladder/HPWHsim/pull/184). 

Further inspection shows that this separate handling of large draws is unneeded. It only leads to different simulation results, compared to the main algorithm, when an inversion occurs, such as when in the inlet temperature exceeds the tank temperature (also uncommon.) This PR removes the large-draw section and applies the same code to all tanks without heat exchangers.

Units testing of large draws was added to `energyBalanceTest.cpp`


nealkruis:
This is great, @spahrenk! As discussed, let's add a comment in the source code to help clarify why we don't need a special case for draws larger than the tank volume, so we don't have this same confusion in the future.

nealkruis (approved):

PR #227, opened 2025-03-12
Make generic update

spahrenk:
Expands capability for refinement of HPWH models. Data structures for specifying model parameters to vary and figure-of-merit targets to meet have been expanded to allow greater flexibility. Currently, only COP polynomial coefficients are fully implemented as parameters, while only energy factors are fully implemented as figures-of-merit. Some simplification was introduced in this update, by inferring the condenser index from the hpwh object, rather than via manual entry.  The lists of parameters and metrics are passed to the `makeGeneric` function in an object of type 'GenericOptions'.

Performance polynomial coefficient entries for Tier-3 and Tier-4 preset models were modified by inter/extrapolation, so that temperatures 50F, 67.46F, and 95F are used as references. This allows independent refinement of the energy factor at each of these temperatures. This method is demonstrated in a unit test added to measureMetrics.cpp.

Functions for refining COP coefficients for specific energy-factor tests E50, UEF, and E95, or the collection of all three, have been added.

Solution by the secant method was added as an alternative to least-squares refinement. The least-squares implementation here remains limited to two parameters and one metric. Secant can be used to refine one parameter to meet one metric.


spahrenk:
@nealkruis I haven't found any info about inlet temperatures for E50 or E95. I see that the test30-test95 in the `test` folder use the temps below:
test30: Ta = 0C, Tinlet = 5C
test50: Ta = 10C, Tinlet = 10C
test70: Ta = 20C, Tinlet = 20C
test95: Ta = 35C, Tinlet = 25C

spahrenk:
@nealkruis Seems to do fine with secant method in unit test, keeping the COP slope fixed. Let me know what else is needed.

src/HPWH.cc
```diff
@@ -5685,6 +5718,12 @@ void HPWH::measureMetrics(FirstHourRating& firstHourRating,
     }
 
     findFirstHourRating(firstHourRating, standardTestOptions);
+    *standardTestOptions.outputStream << "\tFirst-Hour Rating:";
+    *standardTestOptions.outputStream << "\t\tVolume Drawn (L): " << firstHourRating.drawVolume_L
+                                      << "\n";
+    *standardTestOptions.outputStream
+        << "\t\tDesignation: " << FirstHourRating::sDesigMap[firstHourRating.desig] << "\n";
+
```
nealkruis: This kind of messaging should be handled through courier (`send_info`) or produced by the client code.

src/HPWHFitter.hh
```diff
@@ -0,0 +1,314 @@
...
+
+    void leastSquares();
+    void fit();
+};
+
+struct HPWH::GenericOptions
+{
+    std::vector<HPWH::Fitter::MeritInput*> meritInputs;
+    std::vector<HPWH::Fitter::ParamInput*> paramInputs;
+};
```
nealkruis: Please follow [our guidelines](https://github.com/bigladder/atheneum/blob/main/doc/guidelines.md#general-guidelines) and use comments/names to make it easier for me to understand what this struct contains and how it is used. I already mentioned that I don't like "GenericOptions" because it is very vague. Would "FittingParameters" be a more accurate and descriptive name?

src/HPWHFitter.hh
```diff
@@ -0,0 +1,314 @@
...
+#define HPWHFITTER_hh
+
+#include <cfloat>
+
+#include "HPWH.hh"
+
+/**	Optimizer for varying model parameters to match metrics, used by
+ *  make_generic to implement a target UEF. The structure is fairly general, but
+ *  currently limited to one figure-of-merit (UEF) and two parameters (COP coeffs).
+ *  This could be expanded to include other FOMs, such as total energy in 24-h test.
```
nealkruis: Avoid using acronyms, especially when they aren't defined. FOM = figure-of-merit?

src/HPWH.hh
```diff
@@ -1137,9 +1136,25 @@ class HPWH : public Courier::Sender
     {
         bool overrideFirstHourRating = false;
         FirstHourRating::Desig desig = FirstHourRating::Desig::VerySmall;
+
+        bool overrideAmbientT = false;
+        double ambientT_C = 19.7;
+
```
nealkruis: I think we need to reconfigure CustomTestOptions and StandardTestOptions into a "TestConfiguration" that holds:

- Ambient Temperature
- Inlet Temperature

We can then create some static instances of this for the UEF, E50, and E90 tests.

Some options don't ever seem to change, and I don't think we need to include them as something people calling the function can change:

- Setpoint Temperature is always 51.7C
- Change Setpoint? is always false
- Number of Test Thermocouples is always 6

We may still want a separate set of settings that control test output, but really all the file output options should be handled by client code (not in the library itself).

There might be good reasons for the structures that I'm missing, but they don't feel intuitive to me.

src/HPWH.hh
```diff
@@ -1137,9 +1136,25 @@ class HPWH : public Courier::Sender
...
     } customTestOptions;
 
-    void makeGeneric(const double targetUEF);
+    struct Fitter;
+    struct GenericOptions;
+    void makeGeneric(const GenericOptions& genericOptions,
+                     StandardTestOptions& standardTestOptions);
+    void makeGenericUEF(double targetEF, double ambientT_C = 19.7);
+    void makeGenericE50(double targetEF) { return makeGenericUEF(targetEF, 10.); }
+    void makeGenericE95(double targetEF) { return makeGenericUEF(targetEF, 35.); }
```
nealkruis: per comment above, I think we can change these so that they use the static "TestConfiguration"s I described above.

src/HPWH.hh
```diff
@@ -1137,9 +1136,25 @@ class HPWH : public Courier::Sender
...
+    void makeGenericE50_UEF_E95(double targetEF50, double targetUEF, double targetEF95)
```
nealkruis: We still need to figure out what to do if targetEF50 or targetEF95 are not provided. Should we adjust them proportionally to the UEF?

src/HPWH.cc
```diff
@@ -4910,19 +4911,52 @@ void HPWH::initFromFile(string modelName)
...
+        return (1. - ratio) * inletT_C_E50 + ratio * inletT_C_UEF;
+    }
+    if (ambientT_C < ambientT_C_E95)
+    { // linear interp
+        double ratio = (ambientT_C - ambientT_C_UEF) / (ambientT_C_E95 - ambientT_C_UEF);
+        return (1. - ratio) * inletT_C_UEF + ratio * inletT_C_E95;
+    }
+    return inletT_C_E95;
+}
+
```
nealkruis: This really isn't necessary. Each test is a discrete combination of ambient temperature and mains temperature. There are no other tests than those listed in the table provided in the DOE document.

Instead, we can just add additional "TestConfigurations" (see my comment below) if we want to run tests at a different combination of ambient and inlet temperatures.

src/HPWH.cc
```diff
@@ -4910,19 +4911,52 @@ void HPWH::initFromFile(string modelName)
...
     double flowRate_Lper_min = GAL_TO_L(3.);
     if (tankVolume_L < GAL_TO_L(20.))
         flowRate_Lper_min = GAL_TO_L(1.5);
 
-    constexpr double inletT_C = 14.4;   // EERE-2019-BT-TP-0032-0058, p. 40433
-    constexpr double ambientT_C = 19.7; // EERE-2019-BT-TP-0032-0058, p. 40435
-    constexpr double externalT_C = 19.7;
+    const double ambientT_C = (customTestOptions.overrideAmbientT)
+                                  ? customTestOptions.ambientT_C
+                                  : 19.7; // EERE-2019-BT-TP-0032-0058, p. 40435
```
nealkruis: Pull from "TestConfiguration" instead (see comment below).

src/HPWH.cc
```diff
@@ -4910,19 +4911,52 @@ void HPWH::initFromFile(string modelName)
...
+    const double externalT_C = ambientT_C;
+    double inletT_C = findInletT_C(ambientT_C);
```
nealkruis: Pull from "TestConfiguration" instead (see comment below).

src/HPWH.cc
```diff
@@ -5170,9 +5201,12 @@ void HPWH::run24hrTest(const FirstHourRating firstHourRating,
     auto firstDrawClusterSize = firstDrawClusterSizes[firstHourRating.desig];
     DrawPattern& drawPattern = drawPatterns[firstHourRating.desig];
 
-    constexpr double inletT_C = 14.4;   // EERE-2019-BT-TP-0032-0058, p. 40433
-    constexpr double ambientT_C = 19.7; // EERE-2019-BT-TP-0032-0058, p. 40435
-    constexpr double externalT_C = 19.7;
+    const double ambientT_C = (customTestOptions.overrideAmbientT)
+                                  ? customTestOptions.ambientT_C
+                                  : 19.7; // EERE-2019-BT-TP-0032-0058, p. 40435
```
nealkruis: Pull from "TestConfiguration" instead (see comment below).

src/HPWH.cc
```diff
@@ -5170,9 +5201,12 @@ void HPWH::run24hrTest(const FirstHourRating firstHourRating,
...
+    const double externalT_C = ambientT_C;
+    double inletT_C = findInletT_C(ambientT_C);
```
nealkruis: Pull from "TestConfiguration" instead (see comment below).

src/HPWH.cc
```diff
@@ -4988,19 +5022,21 @@ void HPWH::prepForTest(StandardTestOptions& testOptions)
...
         flowRate_Lper_min = GAL_TO_L(1.5);
 
-    constexpr double inletT_C = 14.4;   // EERE-2019-BT-TP-0032-0058, p. 40433
-    constexpr double ambientT_C = 19.7; // EERE-2019-BT-TP-0032-0058, p. 40435
-    constexpr double externalT_C = 19.7;
+    const double ambientT_C = (customTestOptions.overrideAmbientT)
+                                  ? customTestOptions.ambientT_C
+                                  : 19.7; // EERE-2019-BT-TP-0032-0058, p. 40435
+    const double externalT_C = ambientT_C;
+    double inletT_C = findInletT_C(ambientT_C);
```
nealkruis: The First Hour Rating is always performed at 67.5F ambient and 58F inlet temperatures. It should be run independently of the 24 hour test, and the 24 hour test should take the first hour rating draw designation as an input.

nealkruis (changes requested):
@spahrenk this is really good work on addressing the optimization problem!

As I familiarize myself with the DOE test standard, I think there are some ways to streamline the structure of the data and functions.

Finally, there are a number of areas where names/comments can be improved to make it easier for others to understand the code.

Let me know if you have any questions!

src/HPWH.cc
spahrenk: This is very useful, thx.

src/HPWH.cc
spahrenk: The output is directed to a text file here, when an output file path is specified, for the plotting utility. (I replaced that with JSON in a more recent version, and would like to pass the output over a websocket or similar, but haven't figured that out yet.) This was mostly a test feature, so I can definitely direct the output here to courier, rather than the console, when an output file path is not specified.

src/HPWH.hh
spahrenk: Yes.

src/HPWHFitter.hh
spahrenk: Sure, the structure needs some polishing, too.

src/HPWH.cc
spahrenk: @nealkruis 
-Changed parameter names. 
-Defined TestConfiguration. Declared static instances.
-Improved comments. 
-Improved fit initialization. 
-Use UEF temps for first-hour rating. 
-Reported output with courier.

src/HPWHFitter.hh
```diff
@@ -0,0 +1,226 @@
...
+                 std::shared_ptr<Courier::Courier> courier,
+                 HPWH* hpwh_in = nullptr)
+            : Param(courier), hpwh(hpwh_in), tempIndex(tempIndex_in), exponent(exponent_in)
+        {
+        }
+        PerfCoef(PerfCoef& perfCoef, HPWH* hpwh_in = nullptr)
+            : PerfCoef(perfCoef.tempIndex, perfCoef.exponent, perfCoef.courier, hpwh_in)
+        {
+        }
+        enum class PerfCoefType
```
tanaya-mankad: This looks unused.

src/HPWHFitter.hh
```diff
@@ -0,0 +1,226 @@
...
+        {
+            none,
+            PinCoef,
+            COP_Coef
+        };
+        ParamType paramType() override { return ParamType::PerfCoef; }
```
tanaya-mankad: paramType and perfCoeffType seem unused?

src/HPWHFitter.hh
```diff
@@ -0,0 +1,226 @@
...
+        virtual void setValue(double x) = 0;
+        virtual double getValue() = 0;
+
+        virtual void show() {}
+    };
+
+    /// performance coefficient
+    struct PerfCoef : public Param
+    {
+        HPWH* hpwh;
```
tanaya-mankad, line 33: Member variables should almost always be private. Also, is there a design choice here to have nested public structs inside a public struct? Normally classes defined inside other classes are for implementation hiding, so this paradigm looks like an anti-pattern to me.

src/HPWHFitter.hh
```diff
@@ -0,0 +1,226 @@
...
+{
+    ///	base class for variational parameters
+    struct Param : public Sender
+    {
+        double dVal; // differential increment (least squares)
+        Param(std::shared_ptr<Courier::Courier> courier)
+            : Sender("ParamInput", "paramInput", courier)
+        {
+        }
+        Param() : dVal(1.e3) {}
```
tanaya-mankad: Can we call this something more descriptive?

src/HPWHFitter.hh
```diff
@@ -0,0 +1,226 @@
...
+            PinCoef,
+            COP_Coef
+        };
+        ParamType paramType() override { return ParamType::PerfCoef; }
+        virtual PerfCoefType perfCoefType() { return PerfCoefType::none; }
+
+        virtual std::vector<double>& getCoeffs(HPWH::HeatSource::PerfPoint& perfPoint) = 0;
+
+        /// check validity and return reference to HPWH member variable
+        double* getPerfCoeff()
```
tanaya-mankad: I'm having trouble with this function on a couple of levels.
Returning a bare pointer to an internal data member seems ill advised and circumvents const-correctness. Getters should be const, setters not.
Also, it bothers me that there's a function in Param that gets and sets variables in a completely different class! Shouldn't HeatSource be in charge of its own data members?

src/HPWHFitter.hh
```diff
@@ -0,0 +1,226 @@
...
+        /// get difference ratio
+        void evalDiff(double& diff) override
+        {
+            eval();
+            diff = (currVal - targetVal) / tolVal;
+        }
+    };
+
+    /// metric and parameter data retained as shared pts
+    std::vector<std::shared_ptr<Fitter::Metric>> pMetrics;
```
tanaya-mankad: Should be private.

src/HPWHFitter.hh
```diff
@@ -0,0 +1,226 @@
+#ifndef HPWHFITTER_hh
+#define HPWHFITTER_hh
+
+#include <cfloat>
+
+#include "HPWH.hh"
```
tanaya-mankad, line 4: This is a huge header and unnecessary dependency! How about using forward declarations and move any functions that need a full definition to the HPWHFitter.cc file?

src/HPWHFitter.cc
```diff
@@ -0,0 +1,337 @@
...
+
+#include "HPWH.hh"
+#include "HPWHFitter.hh"
+
+//-----------------------------------------------------------------------------
+///	@brief	target function used by secant
+//-----------------------------------------------------------------------------
+static double targetFunc(void* p0, double& x)
+{ // provide single function value
+    HPWH::Fitter* fitter = (HPWH::Fitter*)p0;
```
tanaya-mankad: Use static_cast, not C-style casts.

src/HPWHFitter.cc
```diff
@@ -0,0 +1,337 @@
...
+        metric->eval();
+        return metric->currVal;
+    }
+    return 1.e12;
+}
+
+//-----------------------------------------------------------------------------
+///	@brief	secant function, directly from cse/nummeth.cpp
+//-----------------------------------------------------------------------------
+int secant( // find x given f(x) (secant method)
```
tanaya-mankad: Any reason this isn't also static? 
Or you could consider moving all the statics to an anonymous namespace instead.

src/HPWHFitter.cc
```diff
@@ -0,0 +1,337 @@
...
+    invMatV[1] = (-a * C + b * A) / det;
+    return true;
+}
+
+//-----------------------------------------------------------------------------
+///	@brief	Least-squares minimization (one metric, two parameters)
+/// @note	see [Numerical Recipes, Ch. 15.5](https://numerical.recipes/book.html)
+///         This can be generalized with suitable matrix utilities.
+//-----------------------------------------------------------------------------
+void HPWH::Fitter::leastSquares()
```
tanaya-mankad: It'd be nice if this function had an "action" name; what it _does_.

src/HPWH.cc
tanaya-mankad, line 1: Line 2262 and possibly others: doesn't look like some functions are used anymore. 
Line 2271 and possibly others: Return the object you're "getting" instead of a success code. You're not checking the return value anyway.

tanaya-mankad (changes requested):

src/HPWHFitter.hh
spahrenk: Sure, they are vestigial. I'll take them out.

src/HPWHFitter.hh
nealkruis: I believe this is the whole idea of the fitter. The fitter needs to have a pointer to the data member in HPWH that it is varying as a parameter in the fit so that the fitter can change its value in the optimization process.

That said, I agree that this function should be part of the HPWH class. Ideally, the "fitter" should be completely decoupled from the HPWH class. In theory this should be able to be done by passing in pointers to the parameters and a function that calculates the metric?

src/HPWHpresets.cc
```diff
@@ -3957,17 +3957,21 @@ void HPWH::initPreset(MODELS presetNum)
...
+            {187.064124, 1.939747, 0.0},  // Input Power Coefficients
+            {5.22288834, -0.0243008, 0.0} // COP Coefficients
         });
 
         compressor->perfMap.push_back({
-            70,                        // Temperature (T_F)
-            {148.0418, 2.553291, 0.0}, // Input Power Coefficients (inputPower_coeffs)
-            //{7.207307, -0.0335265, 0.0} // COP Coefficients (COP_coeffs)
-            {6.84694165, -0.0335265, 0.0} // COP Coefficients (COP_coeffs)
+            67.46,                          // Temperature (F)
```
nealkruis: Curious. Why is this 67.46 and not 67.5?

src/HPWHFitter.hh
spahrenk: I think the whole distinction of struct is public membership. There is a need to expose the `dVal` from Param to inherited classes.

src/HPWH.cc
nealkruis: @spahrenk this still needs to be addressed.

src/HPWH.cc
```diff
@@ -4988,20 +5020,19 @@ void HPWH::prepForTest(StandardTestOptions& testOptions)
 
 //-----------------------------------------------------------------------------
 ///	@brief	Find first-hour rating designation for 24-hr test
+/// @note	see EERE-2019-BT-TP-0032-0058, p. 40479 (5.3.3)
 /// @param[out] firstHourRating	    contains first-hour rating designation
 ///	@param[in]	setpointT_C		    setpoint temperature (optional)
-/// @return	true (success), false (failure).
 //-----------------------------------------------------------------------------
-void HPWH::findFirstHourRating(FirstHourRating& firstHourRating, StandardTestOptions& testOptions)
+void HPWH::findFirstHourRating(FirstHourRating& firstHourRating, TestOptions& testOptions)
```
nealkruis: I don't like this pattern. I know it's been this way for a while, but where possible a method like this should return the output (rather than passing it in by reference).

nealkruis: Also, I don't think we should need to pass in TestOptions since it is always the UEF Test configuration.

src/HPWH.cc
```diff
@@ -5145,40 +5175,26 @@ void HPWH::findFirstHourRating(FirstHourRating& firstHourRating, StandardTestOpt
...
+/// @note	see https://www.regulations.gov/document/EERE-2019-BT-TP-0032-0058 (5.4.2)
+/// @param[in/out] testOptions          reporting and configuration
 /// @param[out] testSummary	            contains test metrics on output
-///	@param[in]	setpointT_C		        setpoint temperature (optional)
-/// @return	true (success), false (failure).
 //-----------------------------------------------------------------------------
-void HPWH::run24hrTest(const FirstHourRating firstHourRating,
-                       StandardTestSummary& testSummary,
-                       StandardTestOptions& testOptions)
+void HPWH::run24hrTest(TestOptions& testOptions, TestSummary& testSummary)
```
nealkruis: Similarly, can we return the TestSummary instead of passing it in as a reference?

src/HPWH.cc
```diff
@@ -5659,416 +5675,170 @@ void HPWH::run24hrTest(const FirstHourRating firstHourRating,
     }
 }
 
-void HPWH::measureMetrics(FirstHourRating& firstHourRating,
-                          StandardTestOptions& standardTestOptions,
-                          StandardTestSummary& standardTestSummary)
+//-----------------------------------------------------------------------------
+///	@brief	Modify *this hpwh to meet target metrics
+//-----------------------------------------------------------------------------
+void HPWH::makeGenericE50_UEF_E95(double targetE50,
```
nealkruis: We still need to figure out what to do if targetEF50 or targetEF95 are not provided. Should we adjust them proportionally to the UEF?

src/HPWH.cc
```diff
@@ -5659,416 +5675,170 @@ void HPWH::run24hrTest(const FirstHourRating firstHourRating,
...
+}
 
-    run24hrTest(firstHourRating, standardTestSummary, standardTestOptions);
+//-----------------------------------------------------------------------------
+///	@brief	first-hour rating as a verbose string
+//-----------------------------------------------------------------------------
+std::string HPWH::FirstHourRating::report()
+{
+    std::string results = "";
+    results.append("\tFirst-Hour Rating:\n");
```
nealkruis: Why not just initialize it with this string? The append operation doesn't seem necessary.

src/HPWH.cc
```diff
@@ -5659,416 +5675,170 @@ void HPWH::run24hrTest(const FirstHourRating firstHourRating,
...
 
-    *standardTestOptions.outputStream << "\t24-Hour Test Results:\n";
-    if (!standardTestSummary.qualifies)
+//-----------------------------------------------------------------------------
+///	@brief	24-hr test summary as a verbose string
+//-----------------------------------------------------------------------------
+std::string HPWH::TestSummary::report()
+{
+    std::string results = "";
+    results.append("\t24-Hour Test Results:\n");
```
nealkruis: Same here.

src/HPWH.cc
```diff
@@ -5659,416 +5675,170 @@ void HPWH::run24hrTest(const FirstHourRating firstHourRating,
...
-        standardTestOptions.outputFile.close();
-    }
+    return results;
 }
 
-void HPWH::makeGeneric(const double targetUEF)
+//-----------------------------------------------------------------------------
+///	@brief	Measure the EF and other results of standard 24-hr tests
+//-----------------------------------------------------------------------------
+void HPWH::measureMetrics(TestOptions& testOptions, TestSummary& testSummary)
```
nealkruis: 1. Instead of the HPWH class writing to file, can the TestSummary include the OutputData that can easily processed to a CSV file externally (e.g., in the CLI executable)?
2. The name of this method isn't very accurate. The method writes the 24 hour test time series data to CSV file.

src/HPWH.hh
```diff
@@ -1052,38 +1052,53 @@ class HPWH : public Courier::Sender
...
-        std::ostream* outputStream = &std::cout;
-        bool changeSetpoint = false;
+        std::ostream* resultsStream = NULL;
         std::ofstream outputFile;
-        int nTestTCouples = 6;
+        FirstHourRating::Desig desig = FirstHourRating::Desig::VerySmall;
+        TestConfiguration testConfiguration = testConfiguration_UEF;
+        bool changeSetpoint = true;
         double setpointT_C = 51.7;
+        int nTestTCouples = 6;
```
nealkruis: Per my previous review:

Some options don't ever seem to change, and I don't think we need to include them as something people calling the function can change:

Setpoint Temperature is always 51.7C
Change Setpoint? is always false
Number of Test Thermocouples is always 6

We may still want a separate set of settings that control test output, but really all the file output options should be handled by client code (not in the library itself).

I am wondering if we even need this additional layer. If we use just TestConfiguration, and add the first hour rating designation as an input argument for run24hrTest, we should be able to eliminate the rest of these data members.

src/HPWH.hh
```diff
@@ -1052,38 +1052,53 @@ class HPWH : public Courier::Sender
...
-                             StandardTestOptions& standardTestOptions);
+    void findFirstHourRating(FirstHourRating& firstHourRating, TestOptions& testOptions);
 
     /// run 24-hr draw pattern and compute metrics
-    void run24hrTest(const FirstHourRating firstHourRating,
-                     StandardTestSummary& standardTestSummary,
-                     StandardTestOptions& standardTestOptions);
+    void run24hrTest(TestOptions& testOptions, TestSummary& testSummary);
+
+    static double findInletT_C(double ambientT_C);
```
nealkruis: Remove per earlier comment.

src/HPWHFitter.hh
```diff
@@ -0,0 +1,226 @@
...
+
+        enum class ParamType
+        {
+            none,
+            PerfCoef
+        };
+        virtual ParamType paramType() { return ParamType::none; }
+
+        virtual void setValue(double x) = 0;
+        virtual double getValue() = 0;
```
nealkruis: Question: Do we even need these methods? Can we get away with just a pointer to a double?

src/HPWHFitter.hh
spahrenk: Yes.

src/HPWHFitter.hh
spahrenk: OK.

src/HPWHFitter.cc
spahrenk: Will fix.

src/HPWHFitter.hh
```diff
@@ -0,0 +1,226 @@
...
+        /// get current EF
+        void eval() override
+        {
+            static HPWH::TestSummary testSummary;
+            hpwh->run24hrTest(*testOptions, testSummary);
+            currVal = testSummary.EF;
+        }
+
+        /// get difference ratio
+        void evalDiff(double& diff) override
```
nealkruis: Return `diff` (or better yet `difference_ratio`) instead of passing by reference.

src/HPWH.cc
```diff
@@ -5659,416 +5675,170 @@ void HPWH::run24hrTest(const FirstHourRating firstHourRating,
...
-    CopCoef copT2const(copT2constInfo);
-    CopCoef copT2lin(copT2linInfo);
-
-    pParams.push_back(&copT2const);
-    pParams.push_back(&copT2lin);
+//-----------------------------------------------------------------------------
+///	@brief	Make a generic model with target EF from the current model
+/// by varying COP coef's
+//-----------------------------------------------------------------------------
+void HPWH::makeGenericEF(double targetEF, HPWH::TestOptions& testOptions)
```
nealkruis: I'm wondering if we can get this to return the TestSummary of the final solution. With that (and the idea of including OutputData in the TestSummary, you wouldn't have to re-run the test just to get the data to write out to a file.

nealkruis (changes requested):
Some good progress here! A few items still need to be addressed from my last round of comments and we need to push a bit more to clean up the way we run different tests.

Let me know if you have any questions!

src/HPWHFitter.hh
spahrenk: The objective here is literally to vary the performance coefficients in the HeatSource, possibly other member data, but this could be done through get/set functions. I was originally aiming for performance, so it only retained a pointer to the data element. This at least does some checking first. 

The only reason for the pointer in this function was to share the validity checking for getValue() and setValue(), but I'm sure I can find a better way to do that.

src/HPWHpresets.cc
spahrenk: I mentioned this earlier. It is to decouple the E50 and UEF optimizations. We have been running the UEF test at 19.7 degC = 67.46 degC. We can change the test temperature to 67.5 degC. Alternatively, we could vary a weighted sum of the 50 and 67.5 degF coefficients.

src/HPWH.cc
spahrenk: As a return value? OK.

src/HPWHpresets.cc
nealkruis: The UEF tests are at 67.5F. Not 67.46F. It is very backwards and confusing to use an imprecise unit conversion to justify this.

src/HPWHFitter.cc
spahrenk: Yes.

src/HPWHFitter.hh
spahrenk: I have improved variable names.

src/HPWH.cc
spahrenk: We can do that. The summary can be retained by the EF_Metric object. Each test returns a summary, and those aren't generally used. Also, it is possible that the last evaluation does not correspond to the optimized solution.

OutputData summarizes a single time step for csv output, not the entire test span.

src/HPWHFitter.hh
spahrenk: done

src/HPWHFitter.hh
spahrenk: The aren't needed per se, and I have mentioned several times that pointers are used in the current main version. But, in general, parameters may not correspond to single member variables (e.g, heating capacity), so the functions were introduced for generality. But a pointer is sufficient for input power and COP.  I will use pointers and remove the functions.

src/HPWHFitter.hh
spahrenk: These have been removed.

src/HPWHFitter.hh
spahrenk: I changed this back to direct pointer access.

src/HPWH.cc
nealkruis: @spahrenk the first comment here still needs to be addressed.

src/HPWHFitter.cc
spahrenk: changed to void `performLeastSquaresMinimization()`.

src/HPWH.cc
spahrenk: Should it return std::vector<TestSummary> for the combined function `makeGenericE50_UEF_E95`?

nealkruis: I don't believe we have a use case for processing the test summaries after `makeGenericE50_UEF_E95` yet, so let's not change it yet.

src/HPWH.cc
```diff
@@ -4988,23 +4994,31 @@ void HPWH::prepForTest(StandardTestOptions& testOptions)
...
+    std::string why;
+    if (isNewSetpointPossible(HPWH::testSetpointT_C, maxAllowedSetpointT_C, why, UNITS_C))
     {
-        setSetpoint(testOptions.setpointT_C, UNITS_C);
+        setSetpoint(HPWH::testSetpointT_C);
+    }
+    else
+    {
+        if (isNewSetpointPossible(maxAllowedSetpointT_C, maxAllowedSetpointT_C, why, UNITS_C))
+            setSetpoint(maxAllowedSetpointT_C);
```
nealkruis: I like that you're leveraging an existing function here, though there are some specters here that make it hard to understand what is happening. The design of `isNewSetpointPossible` isn't great, but it can be improved in a separate branch/PR.

nealkruis (approved):

src/HPWH.cc
spahrenk: We just need to assign a setpoint somewhere. Rewrite the section without using the functions?

PR #234, opened 2025-04-25
Remove all file input.

spahrenk:
Removes all use of the original File format.
HPWHsim has supported a `Preset` and a custom `File` formats, which generally mirrored each other. `JSON` was recently added, based on the [hpwh_data_model](https://github.com/bigladder/hpwh-data-model). The `File` format is customized text and was difficult to expand in support of the data model, so this PR removes it.

`cse` defines `HPWHSIM_ABRIDGED` that defines `HPWHsim`'s `HPWH_ABRIDGED`: 
```
#ifndef HPWH_ABRIDGED
    // void initFromFile(std::string modelName);
    void initFromJSON(std::string sModelName);

    // improvised functions to import File -> JSON -> hpwh_data_model -> HPWHsim
    // void readFileAsJSON(std::string modelName, nlohmann::json& j);
    // void initFromFileJSON(nlohmann::json& j);
#endif
```


spahrenk:
@nealkruis I am starting to build off of this branch for the two-inlet fix, etc. Please take a look and merge into main, when possible. Thanks.

src/hpwh/convert.cpp
```diff
@@ -11,7 +11,7 @@ namespace hpwh_cli
 
 /// measure
 static void convert(const std::string& sSpecType,
-                    const std::string& sModelName,
+                    const std::string& modelName,
```
nealkruis, line 14: 🙏🏻

nealkruis (approved):
It feels like HPWHsim just lost 20 pounds!

PR #235, opened 2025-04-30
Fix cases where a tank has two inlets

spahrenk:
Brings in the patch code used in cse from [two-inlet-fix](https://github.com/bigladder/HPWHsim/tree/two-inlet-fix).

Expands tests to check influences of inlet height and draw temperature (draw volumes fixed).
'Tank::updateNodes' (formerly `updateTankTemps`) draws at most one node volume at a time. Each node transfers a volume (heat) to the outlet (usually at the top node), and/or to the node above, and receives a volume ( usually -heat ) from an inlet connected to that node (usually at the bottom node), and/or from the node below.

The volume drawn from node $i$ into node $i+1$ is reduced by the total inlet volume above the node $i$, so a cumulative total is needed of the volume drawn by each inlet working from the top node down.

HPWHsim is coded for two inlets,  at different heights, and which accept different draw volumes and temperatures. 

The multiple inlet draw sequence could be generalized and improved using a std::vector of inlets, rather than an assumed two inlets (one of which is rarely used.)

### DIFF
--- a/src/HPWH.hh
+++ b/src/HPWH.hh
@@ -394,7 +394,7 @@ class HPWH : public Courier::Sender
                     push_back({height, weight});
                     break;
                 }
-                if (weight != node_distribution[i + 1])
+                if (node_distribution[i] != node_distribution[i + 1])
                 {
                     push_back({height, weight});
                 }
--- a/src/HPWH.hh
+++ b/src/HPWH.hh
@@ -975,18 +975,30 @@ class HPWH : public Courier::Sender
     bool isSoCControlled() const;
 
     /// Checks whether energy is balanced during a simulation step.
-    bool isEnergyBalanced(const double drawVol_L,
+    bool isEnergyBalanced(const double inlet1_drawVol_L,
+                          const double inlet1T_C,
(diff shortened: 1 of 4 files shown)


src/HPWH.hh
```diff
@@ -394,7 +394,7 @@ class HPWH : public Courier::Sender
                     push_back({height, weight});
                     break;
                 }
-                if (weight != node_distribution[i + 1])
+                if (node_distribution[i] != node_distribution[i + 1])
```
nealkruis, line 397: Is this related to this fix?

nealkruis (approved):
@spahrenk, looks good. I was curious about one change, though.

src/HPWH.hh
spahrenk: No, it was a bug I introduced when trying to clean-up the structure. It only affected exports on a few models, by not reducing the weighted distribution to the fewest terms.

spahrenk: @nealkruis No, it was a bug I introduced when trying to clean-up the structure. It only affected exports on a few models, by not reducing the weighted distribution to the fewest terms.

PR #237, opened 2025-05-08
Fix unreachable code in getFrac calculation

spahrenk:
The function `HPWH::TempBasedHeatingLogic::getFractToMeetComparisonExternal()` contained unreachable code, causing a warning/error in Windows. This code was added during conversion to weighted distributions and reproduced the original results, but contained and empty fragment. The calculation has been updated using a new method that is more easily interpreted, and also reproduces original results.

A factor of `double nodeDensity = static_cast<double>(hpwh->getNumNodes()) / LOGIC_SIZE;` lingers in the result , but doesn't appear to be needed. Further investigation is underway to understand the intended behavior.

### DIFF
--- a/src/HPWHHeatingLogic.cc
+++ b/src/HPWHHeatingLogic.cc
@@ -261,9 +261,9 @@ double HPWH::TempBasedHeatingLogic::getFractToMeetComparisonExternal()
                 if (firstNode == -1)
                     firstNode = i;
                 calcNode = i;
+                sum += w * hpwh->tank->nodeTs_C[i];
+                totWeight += w;
             }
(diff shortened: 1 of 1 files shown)


spahrenk:
@nealkruis I'll address the node_density issue in a separate branch. This takes care of the Windows warning. Thx.

spahrenk:
@nealkruis Should be good to go.

nealkruis:
@spahrenk It's still in draft mode?

spahrenk:
@nealkruis should be ready to merge now.

nealkruis (approved):

PR #240, opened 2025-05-23
Init presets from cbor

spahrenk:
Initialized "Preset" models from embedded CBOR hpwh_data_model representations. JSON files for these models were generated by exporting the legacy Presets and remain in test/models_json. The CBOR versions are implemented from C++ std::arrays of hexadecimal values inserted in header files placed in the build folder during the CMake build.

Condenser performance calculations were reorganized to make use of std::function member variables, rather than member functions.


.github/workflows/build-and-test.yml
```diff
@@ -71,12 +71,10 @@ jobs:
...
-      - name: Setup python
-        uses: actions/setup-python@v5
-        with:
-          python-version: "3.10"
-      - name: Install Poetry
-        run: pipx install poetry==1.8.1 --python python3.10
+      - name: Install uv
+        uses: astral-sh/setup-uv@v5
+      - name: Set up Python
+        run: uv python install
```
nealkruis: See setup in [atheneum forge](https://github.com/bigladder/atheneum-forge/blob/f4323a1407a3851dba28a641054924b46195db29/.github/workflows/build-and-test.yaml#L24-L27).

scripts/python/data_model/main.py
nealkruis, line 1: What's this file about?

scripts/python/plotting/main.py
nealkruis, line 1: And this file?

src/Condenser.cc
```diff
@@ -589,48 +531,22 @@ void HPWH::Condenser::to(
...
-
-        auto& lookup_vars = map.lookup_variables;
-        checkTo(tempGridValues[0], lookup_vars.input_power_is_set, lookup_vars.input_power);
-        checkTo(
-            tempGridValues[1], lookup_vars.heating_capacity_is_set, lookup_vars.heating_capacity);
-
-        map.grid_variables_is_set = true;
-        map.lookup_variables_is_set = true;
-        perf.performance_map_is_set = true;
+        send_error("Condenser does not have grid data.\n");
```
nealkruis: This is okay for now, but we still want a way to construct an RGI from the evaluate function if the RGI doesn't exist.

scripts/python/data_model/IHPWH_models.json
nealkruis, line 1: As designed, this file needs to be maintained in conjunction with the `enum MODELS` list in HPWH.hh. It seems like the two data sources are redundant. Can we generate the enum in the presets file instead? Then we can actually use the enum variable in the creation of the ordered map instead of a loosely coupled integer?

scripts/python/data_model/incorp_presets.py
nealkruis, line 1: This is a good application for a Jinja template. In general, we should avoid using string concatenation using the `+=` operator. For smaller strings, use f-strings, but for something the size of an entire file you should use Jinja templates.

scripts/python/plotting/pyproject.toml
```diff
@@ -1,16 +1,17 @@
...
-plotly = "*"
-koozie = "*"
-dimes = { git = "https://github.com/bigladder/dimes" }
+[tool.uv]
 
-[build-system]
-requires = ["poetry-core>=1.0.0"]
-build-backend = "poetry.core.masonry.api"
+[tool.uv.sources]
+dimes = { git = "https://github.com/bigladder/dimes#1b8082f06909d551b36e7b784816d7abd915b58a" }
```
nealkruis: Is there a reason this is pegged at such a specific SHA?

src/hpwh-data-model.hh
```diff
@@ -11,11 +11,20 @@
...
+class QuietCourier : public Courier::DefaultCourier
+{
+  public:
+    static inline std::vector<std::string> warnings = {};
+
+  protected:
+    void receive_warning(const std::string& message) override { warnings.push_back(message); }
+};
+void init(std::shared_ptr<Courier::Courier> logger_in = std::make_shared<QuietCourier>());
+} // namespace hpwh_data_model
```
nealkruis: We don't want this anymore, now that @tanaya-mankad has fixed the C++ generation in Lattice, right?

src/hpwh/run.cpp
```diff
@@ -397,6 +395,10 @@ void run(const std::string& sSpecType,
             }
         }
 
+        if (i == 228)
```
nealkruis: What does 228 mean here?

src/Condenser.cc
```diff
@@ -891,64 +756,25 @@ void HPWH::Condenser::getCapacity(double externalT_C,
...
-            C_TO_F(externalT_C), C_TO_F(adjOutletT_C), C_TO_F(adjCondenserTemp_C)};
-        btwxtInterp(input_BTUperHr, cop, target);
-        input_BTUperHr = KW_TO_BTUperH(input_BTUperHr);
-    }
-    else
+    // For accounting add the resistance defrost to the input energy
+    if (resDefrostHeatingOn)
     {
-        getCapacityFromMap(externalT_C, adjCondenserTemp_C, adjOutletT_C, input_BTUperHr, cop);
+        performance.inputPower_W += 1000. * resDefrost.inputPwr_kW;
```
nealkruis: Is there a reason to use the magic number instead of the conversion function?

src/Condenser.cc
```diff
@@ -1012,25 +838,20 @@ void HPWH::Condenser::calcHeatDist(std::vector<double>& heatDistribution)
...
     {
-        double tempInput_BTUperHr = 0., tempCap_BTUperHr = 0., temp_cop = 0.;
         double& externalOutletT_C = hpwh->tank->nodeTs_C[externalOutletHeight];
 
         // how much heat is available in remaining time
-        getCapacity(externalT_C, externalOutletT_C, tempInput_BTUperHr, tempCap_BTUperHr, temp_cop);
+        auto tempPerformance = getPerformance(externalT_C, externalOutletT_C);
 
-        double heatingPower_kW = BTUperH_TO_KW(tempCap_BTUperHr);
+        double heatingPower_kW = tempPerformance.outputPower_W / 1000.;
```
nealkruis: Same question here. Why change this?

src/Condenser.cc
```diff
@@ -1252,410 +1067,102 @@ double HPWH::Condenser::addHeatExternalMP(double externalT_C,
...
-    {
-        if ((environmentT_C > F_TO_C(performanceMap[0].T_F)) &&
-            (extrapolationMethod == EXTRAP_NEAREST))
-        {
-            environmentT_C = F_TO_C(performanceMap[0].T_F);
-        }
+    auto is_integrated = (configuration != COIL_CONFIG::CONFIG_EXTERNAL);
+    auto is_Mitsubishi = (hpwh->model == MODELS_MITSUBISHI_QAHV_N136TAU_HPB_SP);
+    auto is_NyleMP =
+        ((MODELS_NyleC60A_MP <= hpwh->model) && (hpwh->model <= MODELS_NyleC250A_C_MP));
```
nealkruis: Add a comment here to explain why these are special cases.  Add an issue to hpwh_data_model to track the need to allow the data model to define interpolation/extrapolation methods.

src/HPWH.hh
```diff
@@ -1334,57 +1423,151 @@ class HPWH : public Courier::Sender
...
+            return [*this](double externalT_C, double heatSourceT_C)
+            { return evaluate(externalT_C, heatSourceT_C); };
+        }
+
+        inline std::function<Performance(double, double)> use() const
+        {
+            return [this](double externalT_C, double heatSourceT_C)
+            { return evaluate(externalT_C, heatSourceT_C); };
+        }
+    };
```
nealkruis, line 1297: I really like this structure! I'm not as sure about having the tier3 and tier4 sets as members of the struct, though. Can we just make them static variables outside of the struct after the struct is defined?

nealkruis (changes requested):
The functionality appears to be complete, but there are some things that should be cleaned up in the code before this can merge.

scripts/python/plotting/main.py
spahrenk: auto-generated during uv conversion, maybe

scripts/python/data_model/main.py
spahrenk: auto-generated during uv conversion, maybe

scripts/python/data_model/IHPWH_models.json
spahrenk: @nealkruis  It would be useful to start with a std::vector of models containing the preset-id (used by cse), the name (also the json filename), and then the cbor data. The id<->name could reside only in the json list (in the test/models_json folder) until build. We can find an efficient way to reproduce the std::unordered_map functionality, (a map of references to the models vector).

src/hpwh/run.cpp
spahrenk: Left over from debugging.

.github/workflows/build-and-test.yml
spahrenk: That version. doesn't seem to install python. This worked here:
```
      - name: Install uv
        uses: astral-sh/setup-uv@v5
      - name: Set up Python
        uses: actions/setup-python@v5
        with:
          python-version: "3.13"
```

src/Condenser.cc
spahrenk: Restored simplified code to convert integrated condenser polys to grid.

src/Condenser.cc
spahrenk: Comments added. Issue created.

src/HPWH.hh
spahrenk: Changed to HPWH:: static members.

scripts/python/data_model/README.md
nealkruis, line 1: Did you intend to populate this file?

scripts/python/data_model/main.py
nealkruis, line 1: Why is this file still here?

scripts/python/plotting/README.md
nealkruis, line 1: Remove or populate.

scripts/python/plotting/main.py
nealkruis, line 1: Remove.

src/Condenser.cc
```diff
@@ -465,21 +399,22 @@ void HPWH::Condenser::from(const hpwh_data_model::rsairtowaterheatpump::RSAIRTOW
...
-    else if (hpwh->model == MODELS_NyleC185A_C_MP)
+    else if (hpwh->model == hpwh_presets::MODELS::NyleC185A_C_MP)
         resDefrost = {7.25, 5.0, 40.0};
-    else if (hpwh->model == MODELS_NyleC250A_C_MP)
+    else if (hpwh->model == hpwh_presets::MODELS::NyleC250A_C_MP)
         resDefrost = {18.0, 5.0, 40.0};
 
-    if ((MODELS_NyleC25A_SP <= hpwh->model) && (hpwh->model <= MODELS_NyleC250A_C_SP))
+    if ((hpwh_presets::MODELS::NyleC25A_SP <= hpwh->model) &&
+        (hpwh->model <= hpwh_presets::MODELS::NyleC250A_C_SP))
```
nealkruis: Need comment and GitHub issue to address this as well. Are we not characterizing the defrost in the data model?

src/Condenser.cc
```diff
@@ -541,92 +479,117 @@ void HPWH::Condenser::to(
...
         checkTo(inputPowers_W, lookup_vars.input_power_is_set, lookup_vars.input_power);
         checkTo(
             heatingCapacities_W, lookup_vars.heating_capacity_is_set, lookup_vars.heating_capacity);
 
+        map.grid_variables_is_set = true;
+        map.lookup_variables_is_set = true;
         perf.performance_map_is_set = true;
     }
-    else // convert to grid
+    else // convert polySet to grid
```
nealkruis: It's not necessarily a polyset, though, right? It's whatever is in the performance function?

scripts/python/data_model/incorp_presets.py
nealkruis, line 1: You did not address my [previous comment](https://github.com/bigladder/HPWHsim/pull/240#discussion_r2150679959) on this file.

scripts/python/plotting/pyproject.toml
nealkruis: @spahrenk you didn't respond to my question here.

src/hpwh-data-model.hh
nealkruis: @spahrenk you did not respond to my question here.

src/HPWH.hh
```diff
@@ -1334,57 +1246,151 @@ class HPWH : public Courier::Sender
...
+
+    struct PerformancePolySet : public std::vector<PerformancePoly>
+    {
+        PerformancePolySet(const std::vector<PerformancePoly>& vect)
+            : std::vector<PerformancePoly>(vect)
+        {
+        }
+
+        static const PerformancePolySet tier3;
+        static const PerformancePolySet tier4;
```
nealkruis: Why are these still declared here?

nealkruis (changes requested):
@spahrenk, you missed a few of my comments in the last round, and I added a couple more. Can you also merge `main` into this branch? We also need to sort out the differences caused in CSE before this can merge.

scripts/python/plotting/pyproject.toml
spahrenk: Should be fixed.

src/HPWH.hh
spahrenk: Not sure, sorry. Will move them.

src/Condenser.cc
spahrenk: No, I had pointed that out several weeks ago. Will create hpwh_data_model issues.

src/Condenser.cc
spahrenk: True.

scripts/python/data_model/incorp_presets.py
spahrenk: I can look into Jinja.

scripts/python/data_model/README.md
spahrenk: README.md's and main.py's should be gone. Had someone been auto-created.

scripts/python/data_model/main.py
spahrenk: Should be gone now.

spahrenk:
@nealkruis These seem to be all resolved except Jinja. Will work on that.

src/Condenser.cc
spahrenk: @nealkruis Several additional modifications are performed in [configure](https://github.com/bigladder/HPWHsim/blob/ec03f3c5382322f5cfe996b249fa6800c62252a6/src/HPWH.cc#L2874).

nealkruis (approved):

PR #243, opened 2025-06-17
Update lattice: use static constexpr std::string_view

spahrenk:
New version uses `static constexpr std::string_view` in hpwh_data_model.


spahrenk:
@nealkruis @tanaya-mankad This seems to work fine in HPWHsim and cse. Haven't checked build size yet.

tanaya-mankad (approved):

nealkruis:
I believe this was already included in another PR.

PR #247, opened 2025-08-25
Use custom command to detect if presets need to be regenerated

nealkruis:


spahrenk (approved):
Works on my machine.

nealkruis:
@spahrenk I think this is all working now. Please feel free to merge if it looks good to you.

spahrenk:
> @spahrenk I think this is all working now. Please feel free to merge if it looks good to you.

@nealkruis The add_custom_command is very cool! Developers may need to know to delete presets.cpp to regenerate the presets headers. Maybe just a comment in the main CMakeLists.txt?

nealkruis:
@spahrenk the idea here is that presets should only be regenerated if a dependent file changes. That is: the developer should never need to wonder if they need to perform a manual step. I don't think I set up all the dependencies yet, but I think it should be possible to automate the rest of the dependency tree if needed.

spahrenk:
> @spahrenk the idea here is that presets should only be regenerated if a dependent file changes. That is: the developer should never need to wonder if they need to perform a manual step. I don't think I set up all the dependencies yet, but I think it should be possible to automate the rest of the dependency tree if needed.

@nealkruis It will also rebuild if the models.json file changes. Should be fine if the models themselves don't change.

nealkruis:
Thanks, @spahrenk! We can always fine-tune the dependency management as-needed.

PR #248, opened 2025-09-19
Add performance fitting

spahrenk:
Improves performance fitting capabilities.

Adds Bradford White Aerotherm G2 models.

Modifies initGeneric to start with UEF2Generic model.


.github/workflows/build-and-test.yml
```diff
@@ -59,10 +59,10 @@ jobs:
             experimental: true
           - os: ubuntu
             os_ver: "22.04"
-            config: Debug
+            config: Release
             coverage: true
-            cc: gcc-10
-            cxx: g++-10
+            cc: gcc-11
+            cxx: g++-11
```
nealkruis: @tanaya-mankad, this needs to be reverted.

nealkruis (changes requested):

scripts/python/params/params.py
nealkruis, line 1: @tanaya-mankad, if this file is dead, we can remove this directory.

test/ref/DHW_YRLY_Preset.csv
nealkruis, line 1: @tanaya-mankad, I think this file was supposed to be deleted. Can you make sure it isn't used any more?

nealkruis (changes requested):

test/ref/DHW_YRLY_Preset.csv
tanaya-mankad: Everything builds without it, so I removed it. Not sure how else to know...

scripts/python/params/params.py
tanaya-mankad: I removed the dependency supports, but still not sure what the script was for so I'll remove it in my next branch (modernize-scripts).

PR #262, opened 2026-02-16
Add commercial models (IceAir, and BradfordWhite tests)

tanaya-mankad:
1. BradfordWhite models were removed from the "large compressor" set of tests, and so their associated reference files also changed.
2. IceAir double-wall models have been added with model names 
  "IceAirccHPWH275D"
  "IceAirccHPWH550D"


scripts/python/plotting/test_index.json
nealkruis, line 1: @tanaya-mankad Do you know how/if this file is used? It seems to be redundant with stuff in tests/CMakeLists.txt.

nealkruis (approved):
I had one side question that can be addressed in a later updated, but this is looking good. Let's get a CSE branch referencing this build, and we can test changes in CBECC.

scripts/python/plotting/test_index.json
tanaya-mankad: No, I don't. I didn't touch it because it (hopefully) has to do with plotting, which I'm putting on the back burner.

PR #263, opened 2026-06-16
Add commercial models phase 3

tanaya-mankad:
Add IceAir MP and AOSmith (EmergeXC) SP presets.


chipbarnaby (approved):
I am not familiar in detail with the HPWH side and thus I have no comments.

PR #264, opened 2026-06-16
Add commercial models phase 4

tanaya-mankad:
Adding Laars, Nyle, and Watts SP central units to presets.

This PR is stacked on #263 .


chipbarnaby (approved):
Limited knowledge about the HPWH side.

One comment: should values like "external_inlet_height = 0.989583" be idealized to 1.0?  Idealization can provide easier testing / checking in some cases -- 6 sig figs is well beyond the accuracy of the overall model.

